Issue #47, opened 2016-02-23
allow getting process number during test


hellmanj:
Hi, we have a limited number of test accounts with which to run our tests.  We'd like to run our tests in parallel with xdist, but can't use the same account in different tests at the same time.  We can choose which account to use in a fixture, but need to know which worker is running to determine which account to use.  

So is there a way to get the worker number?  I see output like 
```
gw0 I / gw1 I
gw0 [53] / gw1 [53]
```
when I start the test session with two workers.  So can I know if I'm `gw0` or `gw1` from within a fixture?

nicoddemus:
You can get the worker id by acessing the `slaveinput` [dictionary](https://github.com/pytest-dev/pytest-xdist/blob/master/xdist/remote.py#L16). Here's a fixture which makes that information available to tests or other fixtures:

```python
@pytest.fixture
def worker_id(request):
    if hasattr(request.config, 'slaveinput'):
        return request.config.slaveinput['slaveid']
    else:
        return 'master'
```

hellmanj:
Perfect, thanks.

nicoddemus:
@hellmanj would you consider contributing a PR with that fixture or a note in the `README`? :grin:

kunchamlokesh:
Hi,

I have issue related to worker_id. In our project while executing mulit process will be launched and for every  process username & password will be allocated randomly so due to this same credentials were assigned to multiple process., due to this we are facing collision issue. 
We can solve this issue by allocating a static credentials to a process  ie.

suppose we launched 4 process using pytest -n 4
now four process were launched named  pw[0], pw[1], pw[2] & pw[3]. 

I would allocate user1 credentails to process1  and user2 credentials to process2 bsed on name  that pw[0] & pw[1]

Here i am unable to retriever the name of the process . Can any please suggest me how to reterive them

By referring the above comments i have i understood we can user worker_id to retrieve but can you please mention in detail

PR #48, opened 2016-02-23
worker_id fixture as shown in #47

hellmanj:


nicoddemus:
Thanks @hellmanj! I will fix the flakes tests and merge it! :grin:

Issue #76, opened 2016-08-03
Internal adjustments for easier integration with IDEs


nicoddemus:
Here are a check-list of things discussed in pytest-dev/pytest#1613 which are actionable:

* [ ] Hook to receive contents from remote in the master.
* [ ] document that gateway.send could be used from the slave in a plugin to talk to the master in such a hook.
* [ ] document better which default pytest hooks are expected to happen in the master/slave.
* [ ] API to know if currently executing in slave or master.

PR #78, opened 2016-08-05
Add env variables to identify workers

nicoddemus:
While we now have the `worker_id` fixture, having environment variables can help in some situations where you don't have easy access to a fixture, for example in some debugging situations.


nicoddemus:
Resolves one of the items in #76

PR #116, opened 2017-01-23
xdist.plugin: Remove unnecessary `type` for --dist

allanlewis:
1e18f25 corrected an issue with the `type` specification for the `--dist`
argument, setting it to `str`. However, this is the default so there's little
value in specifying it.


nicoddemus:
Thanks!

PR #248, opened 2017-11-22
Issue #241 Make copy during serialization of reprtraceback and reprcr…

timyhou:
This resolves the issue 241 which reported that `report.longrepr` stringification was broken when you enable xdist -nX. The issue had to do with the inadvertent mutation of the individual reprtrackback.reprentries.

### DIFF
--- a/testing/test_remote.py
+++ b/testing/test_remote.py
@@ -79,6 +79,21 @@ def test_remoteinitconfig(testdir):
 
 
 class TestReportSerialization:
+    def test_xdist_longrepr_to_str_issue_241(self, testdir):
+        testdir.makepyfile("""
+            import os
+            def test_a(): assert False
+            def test_b(): pass
+        """)
+        testdir.makeconftest("""
+            def pytest_runtest_logreport(report):
+                print(report.longrepr)
+        """)
+        res = testdir.runpytest('-n1', '-s')
+        res.stdout.fnmatch_lines([
+            '*1 failed, 1 passed *'
+        ])
+
     def test_xdist_report_longrepr_reprcrash_130(self, testdir):
         reprec = testdir.inline_runsource("""
                     import py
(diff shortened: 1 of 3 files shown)


xdist/remote.py
```diff
@@ -111,18 +111,18 @@ def pytest_logwarning(self, message, code, nodeid, fslocation):
...
         for entry in reprtraceback['reprentries']:
             entry_data = {
                 'type': type(entry).__name__,
-                'data': entry.__dict__,
+                'data': entry.__dict__.copy(),
             }
             for key, value in entry_data['data'].items():
                 if hasattr(value, '__dict__'):
-                    entry_data['data'][key] = value.__dict__
+                    entry_data['data'][key] = value.__dict__.copy()
```
timyhou: `reprcrash` and `value.__dict__` actually don't need to be copied but I felt including it might safe us from inadvertent mutations down the line. I can pull it of course.

testing/test_remote.py
```diff
@@ -79,6 +79,21 @@ def test_remoteinitconfig(testdir):
 
 
 class TestReportSerialization:
+    def test_xdist_longrepr_to_str_issue_241(self, testdir):
```
timyhou: I would have preferred doing a test which just test that the report was not mutated before and after both serialization and serialization. I couldn't find a elegant python way of doing this so I am open to suggestions.

nicoddemus: It is hard to properly test that because unfortunately that involves a lot of internal pytest structures which are hard to create. I think this test is enough, I verified it reproduces the issue nicely.

xdist/remote.py
nicoddemus: I think that's fine.

nicoddemus:
Merged with `master` to fix the flake errors.

PR #259, opened 2017-12-21
 Drop support for EOL Python 2.6

hugovk:
Pytest and Py have both dropped Python 2.6 as it's been EOL for 4 years.

Also remove some redundant Python 3.2 code (also EOL and looks like it's already dropped), some minor code cleanup, and add a Python version badge.

---

    ```
    ```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ pytestdebug.log
 .tox/
 .cache/
 .eggs/
+.idea/
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ notifications:
   - pytest-commit@python.org
 
 python:
-- '2.6'
 - '2.7'
 - '3.4'
 - '3.5'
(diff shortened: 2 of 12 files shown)


jurko-gospodnetic:
something broken in our bleeding edge code? related tests seem to be failing...

hugovk:
Yep the pytestmaster and pytestfeatures build jobs are also failing on latest master:

https://travis-ci.org/hugovk/pytest-xdist/builds/319594409

nicoddemus:
I suspect this is breaking because the output has changed and now includes the progress indicator in pytest-3.3, so the code which parses the output is not detecting what it intends to.

hugovk:
Thanks, I've created https://github.com/pytest-dev/pytest-xdist/issues/260 for that, as it's unrelated to this PR.

nicoddemus:
Thanks @hugovk. I'm merging this then, and open a PR shortly for #260.

nicoddemus:
@hugovk  `1.21.0` has been released, thanks again for the PR!

PR #263, opened 2017-12-22
Prepare release 1.21.0

nicoddemus:


nicoddemus:
Just as sanity check, I manually uploaded to `test.pypi.org` and the metadata [seems correct](https://test.pypi.org/simple/pytest-xdist/). I suppose (hope) this will also work when we push the tag and Travis uploads it.

cc @takluyver @jaraco

nicoddemus:
The [metadata is correct](view-source:https://pypi.python.org/simple/pytest-xdist/). 👍

PR #306, opened 2018-07-05
Factor out py.std

asottile:


asottile:
hehe, one more: https://github.com/pytest-dev/execnet/pull/79 :D

PR #307, opened 2018-07-06
Grammar fix

scop:

    ```
    ```


nicoddemus:
Thanks!

PR #342, opened 2018-09-24
Fix warnings transfer between workers and master node with pytest >= 3.8

### LINKED ISSUE #341: Handle pytest_warning_captured in workers
Currently the warnings are no longer being transferred from workers to master.

nicoddemus:
Fix #341

### DIFF
--- a/xdist/remote.py
+++ b/xdist/remote.py
@@ -123,6 +123,18 @@ class WorkerInteractor(object):
             fslocation=str(fslocation),
         )
 
+    # the pytest_warning_captured hook was introduced in pytest 3.8
+    if hasattr(_pytest.hookspec, "pytest_warning_captured"):
+
+        def pytest_warning_captured(self, warning_message, when, item):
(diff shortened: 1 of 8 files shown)


xdist/remote.py
```diff
@@ -123,6 +123,18 @@ def pytest_logwarning(self, message, code, nodeid, fslocation):
...
 
+    # the pytest_warning_captured hook was introduced in pytest 3.8
+    if hasattr(_pytest.hookspec, "pytest_warning_captured"):
+
+        def pytest_warning_captured(self, warning_message, when, item):
+            self.sendevent(
+                "warning_captured",
+                warning_message_data=serialize_warning_message(warning_message),
+                when=when,
+                # XXX what to do with "item" parameter? It can't be serialized
```
nicoddemus: 😬 

In retrospect the `item` parameter was a mistake, it should have been `nodeid`. 😓

PR #348, opened 2018-09-28
WIP use multiple threads for setting up workers

dwiel:
In my experiments, this pull request reduces the amount of overhead in setting up workers by about 66%. This is particularly helpful in situations where a large number of cpu cores are available. In the case of hundreds or thousands of workers, this is still quite a bit of overhead, however this change is minimal and reducing overhead further will likely involve significant amount of work.

### DIFF
--- a/changelog/346.feature
+++ b/changelog/346.feature
@@ -1,0 +1,1 @@
+Use threads when setting up workers. This reduces the overhead of setting up workers by around 66% in many cases.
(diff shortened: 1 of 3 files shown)


dwiel:
Looks like there are still some issues with this. Some look easier to address than others. I'll have to take a closer look a bit later

nicoddemus:
Hi @dwiel,

Thanks for the PR. I would like to close it so we don't have many PRs on the queue, but please do open the PR again when you get back to it.

Thanks!

PR #351, opened 2018-09-30
Issue node shutdown on empty workload

cybojanek:
Remote pytest_runtestloop requires at least 2 tests, or a test
and a shutdown command. In case of inter test locking, for
example using a file lock, the tests could deadlock, since the
shutdown command would not be enqueued, and the last test of a
worker would never finish, and would not allow another worker
to make progress.

    ```
    ```

### DIFF
--- a/changelog/351.bugfix
+++ b/changelog/351.bugfix
@@ -1,0 +1,1 @@
+Fix scheduling deadlock in case of inter-test locking.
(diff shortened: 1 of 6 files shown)


cybojanek:
I'm not sure if calling `node.shutdown()` is safe the way I'm doing it, since there's a lot of logic to follow.

PR #375, opened 2018-11-12
Don't show node setup when running quiet

### LINKED ISSUE #373: Don't show worker nodes when pytest is run quiet
I have a manycore machine (32 threads), and when I use pytest-xdist, the output of the worker node status is a bit bothersome due to showing 32 node status messages multiple times.

This is the output of running pytest-xdist with a subset of the mypy test suite (a full screen CMD window on a 4k monitor):
image

The setup messages take up about 90% of the window.

It would be nice if this could be hidden when pytest is run in quiet mode, as it makes the time to print out the test setup info non-trivial (in addition it fills up the screen making scroll back more annoying).

ethanhs:
This should resolve #373. I thought it would be nice to just say "binging up nodes" to signal *something* is happening, without being as verbose as before.


nicoddemus:
Hi @ethanhs,

Thanks for this and sorry for the late reply.

I agree the case you posted in #375 is *extreme* hehe, but perhaps `-q` should honor and not display anything at all.

Here's what we have currently:

```
λ pytest  foo.py -q
..........
10 passed in 0.04 seconds

λ pytest -n3 foo.py -q
gw0 [10] / gw1 [10] / gw2 [10]
..........
10 passed in 3.91 seconds
```

Perhaps `-q` with xdist should honor the same option as pytest and really display nothing at all?

ethanhs:
>Thanks for this and sorry for the late reply.

No problem! I know full well open source happens at the speed of one's free time :)

>really display nothing at all?

Sure I can do this. Do you want `scheduling tests via ...` omitted as well?

nicoddemus:
I think so.

Thanks for the work again!

ethanhs:
I think(?) the test failures are unrelated, but there could also be some spooky action at a distance.

PR #380, opened 2018-11-21
Skip test which needs pytest_namespace(), removed recently

nicoddemus:


nicoddemus:
We probably didn't notice, because I don't get notifications about the `cron` runs from Travis. 🤔

PR #386, opened 2018-12-05
minor: fix error message and wrongly joined/blackified strings

blueyed:


nicoddemus:
It seems we should fix features :/

nicoddemus:
Going ahead and merge this, the failures are addressed in #389

PR #393, opened 2018-12-12
Fix test_remote_usage_prog which started to break with pytest@features

nicoddemus:
Not sure why this started to fail only on `features` (see https://travis-ci.org/pytest-dev/pytest-xdist/builds/467106087?utm_source=github_status&utm_medium=notification), but this seems to fix it in all versions, although I'm not sure it is correct.

### DIFF
--- a/testing/test_remote.py
+++ b/testing/test_remote.py
@@ -1,4 +1,3 @@
-import os
 import py
 import pprint
 import pytest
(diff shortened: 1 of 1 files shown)


blueyed:
1. The test gets skipped when not on features, but only with https://github.com/pytest-dev/pytest/commit/3445eae737e5d5110674ad5aff1f87263c1be089.

2. The test works with `python -m pytest`, but not `pytest`.

I think this fix is correct, assuming that `runpytest_subprocess` uses `python -mpytest` also.

PR #403, opened 2019-01-14
Fix ResourceWarning: unclosed file in setup.py

BoboTiG:
Also uniformize the usage of the `with` context manager.

Signed-off-by: Mickaël Schoentgen <contact@tiger-222.fr>

### DIFF
--- a/testing/test_remote.py
+++ b/testing/test_remote.py
@@ -153,7 +153,6 @@ class TestReportSerialization:
         for i in range(len(a_entries)):
             assert isinstance(rep_entries[i], ReprEntry)
             assert rep_entries[i].lines == a_entries[i].lines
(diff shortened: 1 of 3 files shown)


testing/test_remote.py
```diff
@@ -153,7 +153,6 @@ def test_repr_entry():
         for i in range(len(a_entries)):
             assert isinstance(rep_entries[i], ReprEntry)
             assert rep_entries[i].lines == a_entries[i].lines
-            assert rep_entries[i].localssep == a_entries[i].localssep
```
nicoddemus: @asottile just FYI. 😁

asottile: 😭 😆 🤣

PR #415, opened 2019-02-12
Improve behavior of --numprocesses=auto and --pdb

jirikuncar:
I was getting annoyed by changing `pytest -ff -n=auto ...` command when I wanted to append only `--pdb`. With this PR one can use `-n=auto` and anytime simply append `--pdb` without getting the `UsageError`.

### DIFF
--- a/changelog/415.feature
+++ b/changelog/415.feature
@@ -1,0 +1,1 @@
+Improve behavior of ``--numprocesses=auto`` to work well with ``--pdb`` option.
--- a/testing/test_plugin.py
+++ b/testing/test_plugin.py
@@ -36,6 +36,7 @@ def test_dist_options(testdir):
 
 def test_auto_detect_cpus(testdir, monkeypatch):
     import os
+    from xdist.plugin import pytest_cmdline_main as check_options
 
     if hasattr(os, "sched_getaffinity"):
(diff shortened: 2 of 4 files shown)


jirikuncar:
@nicoddemus is it necessary to fix the AppVeyor failures [1]?

---
[1] https://ci.appveyor.com/project/pytestbot/pytest-xdist/builds/22311072/job/tpjds44asi1uy5cj

nicoddemus:
Hi @jirikuncar,

Yeah, we need to fix it as it is part of our CI. I think this is due to recent changes on how `UsageError` is displayed (might be related to pytest-dev/pytest#4651 which was merged just a few hours before your build failed).

I think it is enough to change:

```python
result.stdout.fnmatch_lines(
    ["usage: pytest.py *", "pytest.py: error: my_usage_error"]
)
```

to:

```python
result.stdout.fnmatch_lines(
    ["*usage: pytest.py *", "pytest.py: error: my_usage_error"]
)
```

I will push a commit to that effect, let's see. 👍

nicoddemus:
Fixed linting, should be OK now. 👍

nicoddemus:
Thanks a ton @jirikuncar for the PR!

blueyed:
Nice, thanks!

PR #418, opened 2019-02-16
add auto_cpu check if n exceeds host cpus

EvanKepner:
I was working with this plugin and noticed if the `n` I provided exceeded the number of CPUs I had it would generate an `INTERNAL ERROR`. This patch modifies the `xdist.plugin.parse_numprocesses` to return the "auto" value if the input string exceeds the number that is detected. 

I can add to the CHANGELOG for towncrier if this is an acceptable patch. It wasn't clear if this was an intentional optional override for other purposes and a bias in my use case. Running `pytest` in a 3.7 environment passes the current tests, and I've added a new one.

### DIFF
--- a/testing/test_plugin.py
+++ b/testing/test_plugin.py
@@ -64,6 +64,17 @@ def test_auto_detect_cpus(testdir, monkeypatch):
     assert config.getoption("numprocesses") == 2
 
 
+def test_parse_numprocesses(monkeypatch):
+    import xdist.plugin
+    from xdist.plugin import parse_numprocesses
+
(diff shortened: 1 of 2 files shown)


blueyed:
What is the "INTERNAL ERROR" you are getting? `pytest -n 12` works for me for example.

EvanKepner:
Thanks for following up, I did some more testing and it appears to be a false positive. The error only occurs for me running with the coverage plugin and only on one of my tests (even if I run as a singular test). I'll close this until I can isolate the behavior further beyond a one-off in my test design.

PR #422, opened 2019-03-21
Use default with usepdb option

blueyed:
This is required with `-p no:debugging`.

Ref: https://github.com/pytest-dev/pytest/pull/4967.


blueyed:
I don't think it is needed.

nicoddemus:
Hmm you are right. 👍

PR #426, opened 2019-03-26
Serialization hooks (requires pytest 4.4)

nicoddemus:
Replaces xdist's own serialization code with pytest's 4.4 serialization hooks.

This also means that the next pytest-xdist release will require `pytest>=4.4`.

### DIFF
--- a/changelog/426.feature.rst
+++ b/changelog/426.feature.rst
@@ -1,0 +1,5 @@
+``pytest-xdist`` now uses the new ``pytest_report_to_serializable``  and ``pytest_report_from_serializable``
+from ``pytest 4.4`` (still experimental). This will make report serialization more reliable and
+extensible.
+
+This also means that ``pytest-xdist`` now also requires ``pytest>=4.4``.
--- a/changelog/426.feature.rst
+++ b/changelog/426.feature.rst
@@ -1,5 +1,5 @@
-``pytest-xdist`` now uses the new ``pytest_report_to_serializable``  and ``pytest_report_from_serializable``
-from ``pytest 4.4`` (still experimental). This will make report serialization more reliable and
+``pytest-xdist`` now uses the new ``pytest_report_to_serializable`` and ``pytest_report_from_serializable``
+hooks from ``pytest 4.4`` (still experimental). This will make report serialization more reliable and
 extensible.
 
-This also means that ``pytest-xdist`` now also requires ``pytest>=4.4``.
+This also means that ``pytest-xdist`` now requires ``pytest>=4.4``.
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
 from setuptools import setup, find_packages
 
-install_requires = ["execnet>=1.1", "pytest>=3.6.0", "pytest-forked", "six"]
+install_requires = ["execnet>=1.1", "pytest>=4.4.0", "pytest-forked", "six"]
 
 
 with open("README.rst") as f:
--- a/testing/test_remote.py
+++ b/testing/test_remote.py
@@ -3,8 +3,7 @@ import pprint
 import pytest
 import sys
 
-from xdist.workermanage import WorkerController, unserialize_report
-from xdist.remote import serialize_report
+from xdist.workermanage import WorkerController
 import execnet
 import marshal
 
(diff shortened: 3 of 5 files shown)


appveyor.yml
```diff
@@ -1,12 +1,12 @@
 environment:
   matrix:
```
nicoddemus: WIP, will be removed before merging

.travis.yml
```diff
@@ -27,26 +27,26 @@ jobs:
...
-      env: TOXENV=py34-pytestlatest
-    - python: "3.5"
-      env: TOXENV=py35-pytestlatest
-    - python: "3.7"
-      env: TOXENV=py37-pytestlatest
-      sudo: required
-      dist: xenial
-    - python: "2.7"
-      env: TOXENV=py27-pytestmaster
+#    - python: '3.6'
```
nicoddemus: WIP, will be removed before merging

setup.py
```diff
@@ -1,6 +1,6 @@
 from setuptools import setup, find_packages
 
-install_requires = ["execnet>=1.1", "pytest>=3.6.0", "pytest-forked", "six"]
+install_requires = ["execnet>=1.1", "pytest-forked", "six"]  # pytest dep removed (WIP)
```
nicoddemus: WIP, will be removed before merging

nicoddemus:
Rebased and ready for review. 👍

nicoddemus:
@blueyed would appreciate a Ness review when you get the chance. :+1:

changelog/426.feature.rst
```diff
@@ -0,0 +1,5 @@
+``pytest-xdist`` now uses the new ``pytest_report_to_serializable``  and ``pytest_report_from_serializable``
```
blueyed:
```suggestion
``pytest-xdist`` now uses the new ``pytest_report_to_serializable`` and ``pytest_report_from_serializable``
```

changelog/426.feature.rst
```diff
@@ -0,0 +1,5 @@
...
+from ``pytest 4.4`` (still experimental). This will make report serialization more reliable and
```
blueyed:
```suggestion
hooks from ``pytest 4.4`` (still experimental). This will make report serialization more reliable and
```

changelog/426.feature.rst
```diff
@@ -0,0 +1,5 @@
...
+extensible.
+
+This also means that ``pytest-xdist`` now also requires ``pytest>=4.4``.
```
blueyed:
```suggestion
This also means that ``pytest-xdist`` now requires ``pytest>=4.4``.
```

nicoddemus:
Thanks!

alexanderscott:
@nicoddemus could this be published as RC instead of latest release?  at least until the required pytest version is stable?  requiring pytest>=4.4 in a minor release version of this pytest plugin and publishing it doesn't seem very sensible when 4.4 is still experimental and the change will likely break a lot of things which don't have this plugin pinned

nicoddemus:
Hi @alexanderscott, 

>  at least until the required pytest version is stable? requiring pytest>=4.4 in a minor release version of this pytest plugin and publishing it doesn't seem very sensible when 4.4 is still experimental and the change will likely break a lot of things which don't have this package pinned

Why do you say 4.4 is still experimental? It is a normal pytest release. What is experimental are the hooks, in the sense that users should use it with caution because we reserve the right to change them later or even remove them completely. Other than that, the hook implementations have been moved quite literally from `xdist` to the core, so they should be as stable as before, no change there.

About pinning, `pytest-xdist 1.28` requires `pytest 4.4`, so if a user is pinning `pytest` to `4.3.1` (for example), only `pytest-xdist 1.27` should get installed as it is the only version compatible with the pin.

Are you having problems or just a general concern? I appreciate the concern btw.

alexanderscott:
thanks for the quick reply @nicoddemus .  was able to fix my problem but was concerned others may experience similar

> About pinning, `pytest-xdist 1.28` requires `pytest 4.4`, so if a user is pinning `pytest` to `4.3.1` (for example), only `pytest-xdist 1.27` should get installed as it is the only version compatible with the pin.

This logic makes sense but isn't what I experienced for some reason.  Had several different versions of pytest pinned to microservices (for ex/ `'pytest~= 4.0.0'`) without pinning `pytest-xdist`, expecting a compatible version would be sensibly installed.  However, builds started failing today with `pluggy.manager.PluginValidationError: Plugin 'xdist' could not be loaded: (pytest 4.0.2 (/mnt/jenkins/workspace/tox-phab-pipe/src/.tox/py27-django16/lib/python2.7/site-packages), Requirement.parse('pytest>=4.4.0'))!` .  Pinning `1.27` fixed the dependency issue.

Might just be an issue specific to my setup (in which case please ignore), but this dependency mismatch hasn't occurred for me before with `xdist` or any other pytest plugins I have configured previously.

nicoddemus:
Hi @alexanderscott,

Thanks for the description.

Hmm the problem is probably pip's dependency solver, which is long standing issue: https://github.com/pypa/pip/issues/988.

Unfortunately our hands are tied for two reasons: if the dependency solver is not doing it's job correctly, then it wouldn't matter in which version we released this change, it would still not honor the version restrictions even if we released pytest-xdist as 2.0.0 or RC. 😞 The other reason is that we can't really change the release once it is out, the best we can do is make another release pulling back this change, but I fear this might even add more confusion to the mix. 

Let's see if more people manifest this problem.

Again, thanks for reaching out, it is important to have this kind of dialog to ensure our expectations are on the same page. 👍

alexanderscott:
No problem, and thanks for the info and your thorough reply @nicoddemus .  Keep up the good work!

PR #454, opened 2019-07-24
broadcast child shouldstop or fail Fixes pytest-dev/pytest#5655

graingert:

    ```
    ```

### DIFF
--- a/src/xdist/dsession.py
+++ b/src/xdist/dsession.py
@@ -254,12 +254,19 @@ class DSession(object):
         self.config.hook.pytest_runtest_logreport(report=rep)
         self._handlefailures(rep)
 
-    def worker_runtest_protocol_complete(self, node, item_index, duration):
+    def worker_runtest_protocol_complete(
+        self, node, item_index, duration, shouldfail, shouldstop
+    ):
         """
         Emitted when a node fires the 'runtest_protocol_complete' event,
         signalling that a test has completed the runtestprotocol and should be
         removed from the pending list in the scheduler.
         """
+        if shouldfail:
+            self.shouldfail = shouldfail
+        if shouldstop:
+            self.shouldstop = shouldstop
+
         self.sched.mark_test_complete(node, item_index, duration)
 
     def worker_collectreport(self, node, rep):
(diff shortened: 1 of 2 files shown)


graingert:
@nicoddemus I tested this locally with my testcase here: https://github.com/pytest-dev/pytest/issues/5655

graingert:
@nicoddemus I'm not sure what sort of test I should add here though

graingert:
it also works for me on the big project I'm testing it on - not sure how to handle multiple simultaneous test failures though

nicoddemus:
Hi @graingert,

Sorry for the delay!

I think an integration test with stepwise, like described in in #458 (I've moved it from `pytest` to this repository) should be enough.

Also please add a `458.bugfix.rst` CHANGELOG entry, something like:

```
The ``--stepwise`` option is now supported correctly.
```

graingert:
It's not actually correct though, because when multiple tests fail the suite starts from a random test

nicoddemus:
> It's not actually correct though

You mean the changelog message or the actual behavior being observed?

graingert:
The new behaviour with this PR. It makes stepwise mostly workable but a bit
frustrating

nicoddemus:
Definitely, but I'm not sure if we can improve that.

graingert:
`-x` works perfectly however as it operates directly on the master process.
Is there a way for plugins to hook into DSession?

nicoddemus:
> Is there a way for plugins to hook into DSession?

Not right now, but we can include that depending on what's needed.

graingert:
@nicoddemus having worked with this for a while, it's actually extremely frustrating! Quite often failing tests are skipped over, or not run and skipped over the next session

nicoddemus:
> @nicoddemus having worked with this for a while, it's actually extremely frustrating! Quite often failing tests are skipped over, or not run and skipped over the next session

Sorry to hear. Does this PR makes things better for you?

graingert:
Sorry I meant that this pr takes stepwise+xdist from not working at all to a frustrating half-working atate

nicoddemus:
Ahh OK, thanks for the clarification. 👍

graingert:
Is there a way to make xdist wait for all currently executing tests before
stopping at shouldstop

nicoddemus:
Closing this as stale, but feel free to reopen if you want to tackle this again @graingert!

PR #473, opened 2019-10-02
Noisy rsync

cmecklenborg:

    ```
    ```


nicoddemus:
Ignoring the py38-windows failures because they should be fixed in the next pytest release and are not related to pytest-xdist.

nicoddemus:
Hey @cmecklenborg, thanks for the PR and we're terribly sorry for the silence here, this must have fell through the cracks.

cmecklenborg:
@nicoddemus Thanks for the merge! Sorry for not following up on this sooner myself. :grimacing:

nicoddemus:
Not at all, we dropped the ball here. But thanks again!

PR #483, opened 2019-11-02
Add an example on how to share session fixture data to README

nicoddemus:
As discussed in #385

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -95,6 +95,59 @@ any guaranteed order, but you can control this with these options:
   in version ``1.21``.
 
 
+Making session-scoped fixtures execute only once
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``pytest-xdist`` is designed so that each worker process will perform its own collection and execute
+a subset of all tests. This means that tests in different processes requesting a high-level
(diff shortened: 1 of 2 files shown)


larsblumberg:
Great to have this MR! I want to throw in code that we're using in your product that makes sure that only worker sets up the database. Maybe this MR can profit from our solution which is slightly less complex but comes at the expense to use the external library `posix_ipc` for creating a shared Semaphore:

```python
from posix_ipc import O_CREAT, Semaphore

def pytest_configure(config):
    from fahrtwind.models.meta import Base
    # As we run tests in parallel, only let the first test runner create the schema to avoid collision
    # TODO: Write a blog post about this technique
    with Semaphore('/fahrtwind-testing', flags=O_CREAT, initial_value=1):
        engine = create_engine(environ['TESTING_DB_URL'])
        if not engine.table_names():
            Base.metadata.create_all(engine)
```

This method is contained in the root `conftest.py` and is run at the start-up of eacher worker. This approach might be interesting to you if `pytest` accepts to add the external dependency.

nicoddemus:
Hi @larsblumberg,

Thanks sharing! I'm not sure this would make sense in pytest-xdist though as `posix_ipc` is not Windows compatible.

But I think it would be worthwhile to add it as an example next to this one in the README (with the caveat that it only works on Linux), would you like to open a PR?

PR #486, opened 2019-11-09
Add support for Python 3.8

hugovk:

  * Name it `$issue_id.$type` for example `588.bugfix`

    ```
    ```

---

Also rename `appveyor.yml` to `.appveyor.yml`. Let's group it with the other dotfiles for consistency and less clutter.

> AppVeyor supports dot-file-style YAML named `.appveyor.yml` as is. 

https://www.appveyor.com/docs/build-configuration/#yaml-file-alternative-naming


nicoddemus:
Thanks @hugovk!

PR #500, opened 2020-01-19
Custom test grouping and test group order logic

SalmonMode:
It's been long requested (#18) to allow for custom logic for the distribution of tests among test groups in a simple and clean way. This adds that functionality by providing a hook that each tests nodeid is passed to, and whatever is returned by the hook for a given test is that test's test group name. All tests with the same test group name will be batched together and passed to the same worker together.

I also added another hook that grants access to the finished collection of test groups so that the order of the groups can be changed as needed. This allows for certain optimizations to be made, such as putting slow tests in the same group and placing that group in the beginning of the collection in order to speed the test suite up.

I also took the opportunity to clean up the logic quite a bit as there was a _lot_ of repetition between the different load schedulers. Now they are unified into the same base class (except for `EachScheduling`) with the only differing logic being what their default grouping logic is. I felt I should leave `EachScheduling` alone for now, but I imagine it could also use the same base class by simply having every test in the same group, and then duplicating that group N times. But that seemed kinda memory heavy.

- [x ] Make sure to include reasonable tests for your change if necessary

- [x ] We use [towncrier](https://pypi.python.org/pypi/towncrier) for changelog management, so please add a *news* file into the `changelog` folder following these guidelines:

    ```
    ```


nicoddemus:
Hi @SalmonMode, 

Thanks a lot for the PR! As you mention this is a long-requested feature so all the work you have put into it is more than welcome!

I like the approach of providing "low level" hooks that lets users and plugins customize test grouping to be sent to workers. Also so far in the review I'm very satisfied with the code quality and documentation you've put into it. 👍 

I'm writing this just so to let you know that it will probably take awhile to review a PR of this magnitude, so please be patient! 😁

SalmonMode:
@nicoddemus haha I appreciate it :grin: . Take your time. I spent a very long time on it (granted, at least half of that time was on what to name the hooks). 

As a side note, the name I chose was because I figure at some point in the future, this may be expanded to allow grouping based on markers, rather than just names, but I couldn't figure out how to pull those in in a good way, and wanted to get the PR in rather than holding onto it for longer.

ssbarnea:
Any chance to refresh this one?

nicoddemus:
> Any chance to refresh this one?

Thanks for the ping @ssbarnea! Indeed I had lost sight of this.

Will make time for completing my review this week. 👍

skhomuti:
@nicoddemus just reminding about the existence of this PR :)

Wilsontomass:
Heya @nicoddemus, sorry for the bump but we would be extremely grateful for a PR of this kind to be added to pytest-xdist. Is it still feasable? Has some other functionality superseded it?

RonnyPfannschmidt (commented):
This is quite large, so itll be a while before I can take a deeper look

I want to note that it's unfortunate we don't have a good way to transfer details about marks with the nodeids, we should add a hook that allows to inform the coordinator about affinities (session scope fixtures, certain configurations)

nicoddemus:
Thanks a lot for the ping!

I had left a bunch of comments, but forgot to publish them 🤦 :

image

I just published them, I will give it another review later.

nicoddemus:
> we should add a hook that allows to inform the coordinator about affinities (session scope fixtures, certain configurations)

Just so we are in the same page, you mean in this PR or as a follow up?

src/xdist/newhooks.py
```diff
@@ -55,3 +55,55 @@ def pytest_xdist_node_collection_finished(node, ids):
...
+
+    This would then defer to the default distribution logic for any tests this
+    can't apply to (i.e. if this would return ``None`` for a given ``nodeid``).
+    """
+
+@pytest.mark.trylast
+def pytest_xdist_order_test_groups(workqueue):
+    """Sort the queue of test groups to determine the order they will be executed in.
+
+    The ``workqueue`` is an ``OrderedDict`` containing all of the test groups in the
```
nicoddemus, line 103:
```suggestion
    The ``workqueue`` is an ``OrderedDict`` of ``group => list of node ids`` containing all of the test groups in the
```

src/xdist/scheduler/load.py
```diff
@@ -1,286 +1,459 @@
...
+    collection. This is used to know when all nodes have finished collection.
+
+    Work units can also be considered to be "test groups", as the tests inside
+    should not be split up across multiple workers, and should be run within
+    the same work cycle for a single worker. By default, this does not attempt
+    to group the tests in any way, so each work unit would only contain a
+    single test. This is designed to be extensible so that custom grouping
+    logic can be applied either by making a child class from this and
+    overriding the ``get_default_test_group`` method, or by defining the
+    ``pytest_xdist_set_test_group_from_nodeid``.hook. If the hook is used, but it returns
```
nicoddemus, line 31:
```suggestion
    ``pytest_xdist_set_test_group_from_nodeid`` hook. If the hook is used, but it returns
```

src/xdist/scheduler/load.py
```diff
@@ -1,286 +1,459 @@
...
+                (...)
+            }
+
+    :assigned_work: Ordered dictionary that maps worker nodes with their
+       assigned work units.
+
+       ::
+
+            assigned_work = {
+                '<worker node A>': {
```
nicoddemus, line 71:
```suggestion
                'gw0': {
```

Using ids as we have them internally will make this easier to understand I think.

src/xdist/scheduler/load.py
```diff
@@ -1,286 +1,459 @@
...
+                (...)
+            }
+
+    :registered_collections: Ordered dictionary that maps worker nodes with
+       their collection of tests gathered during test discovery.
+
+       ::
+
+            registered_collections = {
+                '<worker node A>': [
```
nicoddemus, line 88:
```suggestion
                'gw0': [
```

src/xdist/newhooks.py
```diff
@@ -55,3 +55,55 @@ def pytest_xdist_node_collection_finished(node, ids):
 @pytest.mark.firstresult
 def pytest_xdist_make_scheduler(config, log):
     """ return a node scheduler implementation """
+
+
+@pytest.mark.trylast
```
nicoddemus, line 60:
```suggestion
@pytest.hookspec(trylast=True, firstresult=True)
```

Besides using the new syntax, I think we should add `firstresult` here to make it clear in the API that we will only use the first result from the hook.

src/xdist/newhooks.py
```diff
@@ -55,3 +55,55 @@ def pytest_xdist_node_collection_finished(node, ids):
...
+def pytest_xdist_set_test_group_from_nodeid(nodeid):
```
nicoddemus, line 61:
```suggestion
def pytest_xdist_get_test_group_from_nodeid(nodeid):
```

Perhaps "get" better conveys that we should return the test group? To me "set" conveys that I should *set* the test group somewhere.

src/xdist/scheduler/loadscope.py
```diff
@@ -1,18 +1,14 @@
...
-from _pytest.runner import CollectReport
+from .load import LoadScheduling
 from py.log import Producer
-from xdist.report import report_collection_diff
-from xdist.workermanage import parse_spec_config
 
 
-class LoadScopeScheduling(object):
+class LoadScopeScheduling(LoadScheduling):
     """Implement load scheduling across nodes, but grouping test by scope.
```
nicoddemus, line 6: So `loadfile` and `loadscope` are distribute the tests identically now? Or am I missing something?

nicoddemus (commented):
Reviewed the implementation and it looks good, left a few more comments.

I personally would like to move some of the current code that manipulates workloads/test groups abstracted away into separate classes, instead of having all the details implemented into the scope itself, but definitely this can be done later if someone decides to tackle it.

PR #505, opened 2020-02-18
New methods `is_xdist_worker`, `is_xdist_master`, `get_xdist_worker_id`. 

### LINKED ISSUE #504: Handy API to know information about master/slaves from session
As mentioned in https://github.com/smarie/python-pytest-harvest/issues/36 (and discussed already with @nicoddemus in https://github.com/smarie/python-pytest-harvest/issues/32) it would be great to have an elegant API to answer these simple questions:

 - `get_xdist_worker_id(session)`: would return `session.config.slaveinput['slaveid']` or raise an exception if this not an xdist worker
 - `is_xdist_worker(session)`: would return `True` if this is an xdist worker, `False` otherwise
 - `is_xdist_master(session)`: would return `True` if this is the xdist master, `False` otherwise (**important**: this would return `False` when distribution has not been activated and "normal" pytest has been run on a single process)

Would you like a PR for this ?

When/if this API is done, we should not forget to post the answer on [this SO question](https://stackoverflow.com/questions/60212389/how-can-i-know-from-inside-a-pytest-hook-if-pytest-xdist-is-enabled)

smarie:
Fixes #504 .

I did not create any additional test as I am not sure how/where these should be put.


nicoddemus:
Thanks a lot @smarie and really sorry about the long delay!

smarie:
No worries, thanks for merging !

PR #502, opened 2020-02-11
Move checking of options to pytest_configure

blueyed:
This allows for `pytest -n 2 --pdb` to stop / help with errors
before/during `pytest_configure`.
It also appears to be sensible to check the config in the hook meant for
it, instead of in `pytest_cmdline_main`.

TODO (after approval):

- [ ] changelog

PR #507, opened 2020-02-22
tests: re-harden test_remote_usage_prog

blueyed:
The meaning of the test is to check that "prog" is not "-c", which was
weakened in 93780dccc.

/cc @nicoddemus @jirikuncar


nicoddemus:
Closing as stale.

blueyed:
Why is it stale?

nicoddemus:
https://github.com/pytest-dev/pytest-xdist/pull/502#issuecomment-828653838

PR #531, opened 2020-05-03
Increase pexpect timeouts to prevent tests from failing

mgorny:

    ```
    ```

### DIFF
--- a/testing/acceptance_test.py
+++ b/testing/acceptance_test.py
@@ -332,7 +332,7 @@ class TestDistribution:
                 time.sleep(10)
         """
         )
-        child = testdir.spawn_pytest("-n1 -v")
+        child = testdir.spawn_pytest("-n1 -v", expect_timeout=30.0)
         child.expect(".*test_sleep.*")
         child.kill(2)  # keyboard interrupt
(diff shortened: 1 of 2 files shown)


mgorny:
Do you consider this change worth a changelog entry?

nicoddemus:
Hi @mgorny, thanks for the PR!

Linting failed, I've fixed it and pushed to your branch.

> Do you consider this change worth a changelog entry?

Not necessary, this is an internal change and not visible to users. 👍 

Can you describe why this change is necessary though? I've never seen those tests fail.

mgorny:
To be honest, I've tried three times and got all three tests failing every time (once with PyPy3, twice with CPython 3.8). It might be that my system is slow, background tasks keep it slow-ish or just Portage/sandbox causes unexpected slowness. In any case, increasing the timeout fixed it for me.

I can imagine our arch teams would hit it when testing pytest-xdist on more exotic architectures like ARM.

nicoddemus:
OK, thanks for the explanation.  

I don't see any harm in increasing the timeout, was just curious and wanted to leave it registered here the reasons. 👍

PR #536, opened 2020-05-26
Randomize load scope scheduler

### LINKED ISSUE #229: Run tests in random order on nodes
Sorry if this has been covered elsewhere or is obvious, but I would like to be able to run our test suite with the tests running in a random order where the order is controlled by a random seed to allow reproducibility. At the moment I can do that with `pytest-random-order` or `pytest-randomly` but as soon as I use `-n` everything gets confused because it seems the randomization messes with the test to node allocation and the subprocesses complain that they have been scheduled the same tests or that tests are missing.

Is there any way the xdist scheduler can play nicely with `pytest-random-order` or any way to control the ordering in `xdist` itself?

My motivation for this is that we are finding that some of the tests are sensitive to test execution order and fail if run first or on their own.

Hyzerputt:
Attempt to resolve #229.  I worked through various options to try and make the solution less specific than for just loadscope scheduling but I don't feel up to the task while maintaining python 2.7 compatibility.  Ultimately I tried to follow the patterns in the existing code for adding schedulers.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -87,12 +87,17 @@ any guaranteed order, but you can control this with these options:
   by **class** for *test methods*, then each group will be sent to an available worker,
   guaranteeing that all tests in a group run in the same process. This can be useful if you have
   expensive module-level or class-level fixtures. Currently the groupings can't be customized,
-  with grouping by class takes priority over grouping by module.
+  with grouping by class taking priority over grouping by module.  Use loadscopeshuffled for
+  randomizing the groupings execution order.
   This feature was added in version ``1.19``.
 
 * ``--dist=loadfile``: tests will be grouped by file name, and then will be sent to an available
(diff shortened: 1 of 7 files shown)


Hyzerputt:
The failed tests appear to be an inaccurate test expectation (order of test distribution) with regard to ensuring that #277 is fixed.  Not being fluent in the project, I would appreciate advice on how to better verify #277 for the new scheduler.

Hyzerputt:
I marked the test_module_single_start acceptance xfail for new scheduler while awaiting guidance on how to validate #277 when the test distribution order is randomized.

nicoddemus:
Thanks @Hyzerputt and sorry for the silence on this one.

We are discussing an alternative approach in https://github.com/pytest-dev/pytest-xdist/pull/632.

PR #546, opened 2020-06-26
Add missing dependency on py

swt2c:


nicoddemus:
Thanks @swt2c, but this is not really needed because `pytest` already declares that dependency. 👍

PR #541, opened 2020-06-12
Remove deprecated "slave" aliases

Zac-HD:
We started this process in 2017 with #234, and it's time to complete it.  Xref https://github.com/pytest-dev/pytest/issues/7344

Implementation note: after  is in a released version of pytest, I'll bump the `pytest` required version for `pytest-xdist` and then merge this :slightly_smiling_face: 

- [x] Merge https://github.com/pytest-dev/pytest/pull/7356 for upstream `pytest` support
- [x] Bump minimum dependency version to `pytest >= 6.0.0` (after that release)
- [x] Fix other plugin compatibility issues linked below
  - [x] PRs opened for all plugins
  - [x] PRs merged for all plugins
  - [x] Fix released for all plugins (mostly done, not treating this as a blocker)
- [x] Merge this PR
- [ ] Release `pytest-xdist` 2.0 :tada:

PR #557, opened 2020-07-11
Modernise codebase for Python 3.5+

hugovk:
PR https://github.com/pytest-dev/pytest-xdist/pull/541 drops support for EOL Python 2.7.

This PR **updates https://github.com/pytest-dev/pytest-xdist/pull/541 (not `master`)** to modernise the codebase:

* Set `py35` as the target (ie. minimum) version for Black
* Upgrade Python syntax with `pyupgrade --py3-plus`
* Remove some code for `sys.version_info[0] < 3:`
* Remove the six library

---

- [n/a?] We use [towncrier](https://pypi.python.org/pypi/towncrier) for changelog management, so please add a *news* file into the `changelog` folder following these guidelines:

    ```
    ```


Zac-HD:
Implementation note: I propose to wait for the release of Pytest 6, have a final review of #541 without these additional changes, then merge this PR into 541 and finally merge 541 into master.

Zac-HD:
I've rebased this onto #541 locally, and consider it 'merged' even though GitHub doesn't :roll_eyes: 

Thanks again @hugovk!

PR #558, opened 2020-07-13
Fix rsyncdirs usage with pytest 6.0

nicoddemus:
Fix pytest-dev/pytest#7487


ionelmc:
So I tested the master locally, it does fix it.

PR #561, opened 2020-07-24
Improve description of testing with multiple CPU's

adamchainz:
* Describe ``--numprocesses`` as the long form of ``-n``.
* Describe hidden ``--dist no`` option.
* Remove version annotations (it's been 2.5 years since the most recent mentioned version).
* Consistently use ``--flag value`` rather than ``--flag=value``.
* Drop tease about customization of test grouping.
* General language improvements to better describe how this all works.


adamchainz:
👍

PR #566, opened 2020-07-31
Fix test_config_initialization condition for pytest<5

mgorny:
Fix test_config_initialization to be correctly skipped on pytest<5,
by explicitly checking pytest version (idea copied from
test_looponfail_removed_test).  The current conditions are insufficient
-- the outer check wrongly assumes pytest>=5 will always be used
on Python 3 (which is not true if you need the same version to support
both Python 2 and Python 3), and the inner condition apparently
wrongly assuming that invocation_params attribute is not present
in pytest-4.

----

    ```
    ```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,8 @@ jobs:
       env: TOXENV=py36-pytestlatest
     - python: "3.7"
       env: TOXENV=py37-pytestlatest
+    - python: "3.9-dev"
+      env: TOXENV=py39-pytestlatest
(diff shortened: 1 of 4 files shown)


nicoddemus:
Thanks a lot @mgorny and sorry for the delay!

We just merged #541 which drops support for pytest<6, so the check is no longer really necessary. I've rebased your commits and cleaned up accordingly in #570, which then supersedes this one.

Thanks again!

PR #568, opened 2020-08-03
[compat] Drop Python 2.x supports

Djailla:
As pytest no longer supports Python 2.7, drop it from this package too

    ```
    ```


nicoddemus:
Thanks a lot for the PR @Djailla, but this is superseded by #541, which has been waiting a bit for plugins to catch up before getting merged. 👍  

Thanks again!

PR #572, opened 2020-08-06
Remove compat for pytest < 6

Zac-HD:
More follow-up for #571, removing conditionals which exist for now-unsupported old versions of pytest.


Zac-HD:
Happy to help! Ripping out compatibility shims is always satisfying 😁

PR #575, opened 2020-08-12
Prepare release 2.0.0

nicoddemus:
The py38-windows build will fail due to https://github.com/pytest-dev/pytest/issues/7638, but it is not related to pytest-xdist and will be fixed in the next pytest release.


nicoddemus:
cc @Zac-HD @hugovk 🎉

PR #599, opened 2020-09-10
doc: fix invalid url(socketserver.py)

yili1992:

    ```
    ```


nicoddemus (approved):
Thanks!

nicoddemus:
The failure is unrelated and should be fixed by https://github.com/pytest-dev/pytest/pull/7721.

PR #643, opened 2021-03-20
 Use 'main' to refer to pytest default branch in tox env names.

graingert:

    ```
    ```

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,7 @@ envlist=
 extras = testing
 deps =
   pytestlatest: pytest
-  pytestmain: git+https://github.com/pytest-dev/pytest.git@main
+  pytestmain: git+https://github.com/pytest-dev/pytest.git
 commands=
   pytest {posargs}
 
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -2,14 +2,14 @@
...
-  py38-pytestmaster
+  py38-pytestmain
   py38-psutil
 
 [testenv]
 extras = testing
 deps =
   pytestlatest: pytest
-  pytestmaster: git+https://github.com/pytest-dev/pytest.git@main
+  pytestmain: git+https://github.com/pytest-dev/pytest.git@main
```
graingert, line 12: @nicoddemus doesn't this work without a branch name?

```suggestion
  pytestmain: git+https://github.com/pytest-dev/pytest.git
```

nicoddemus: Probably yes, this might be a left over when switched from having a `features` branch.

nicoddemus:
Thanks!

PR #654, opened 2021-04-25
Drop Python 3.5 support

nicoddemus:
pytest no longer supports it, and our CI is broken because of it.

Also removed `linting` given we are using pre-commit.ci.

Should we run pyupgrade as well?


nicoddemus:
Going ahead and merging this as I think it is simple change. 👍

PR #667, opened 2021-06-03
fix sys.path for local workers  Fixes #421

### LINKED ISSUE #421: xdist adds current directory to sys.path
With 1.26.1.
```python
import sys

def test_path():
    assert "" not in sys.path
```
```
$ pytest test_path.py
<OK>
$ pytest test_path.py -n 1
E     AssertionError....
```
Obviously this can cause problems/change behaviour depending on where you are running the tests from.

Possibly related to #376?

graingert:
#421 

    ```
    ```

### DIFF
--- a/src/xdist/workermanage.py
+++ b/src/xdist/workermanage.py
@@ -9,6 +9,7 @@ import pytest
 import execnet
 
 import xdist.remote
+from xdist.plugin import _sys_path
 
 
 def parse_spec_config(config):
@@ -261,7 +262,8 @@ class WorkerController:
         remote_module = self.config.hook.pytest_xdist_getremotemodule()
         self.channel = self.gateway.remote_exec(remote_module)
         # change sys.path only for remote workers
-        change_sys_path = not self.gateway.spec.popen
+        # restore sys.path from a frozen copy for local workers
+        change_sys_path = _sys_path if self.gateway.spec.popen else None
         self.channel.send((self.workerinput, args, option_dict, change_sys_path))
 
         if self.putevent:
--- a/testing/test_remote.py
+++ b/testing/test_remote.py
@@ -234,7 +234,7 @@ def test_remote_env_vars(testdir):
 
 
 def test_remote_inner_argv(testdir):
-    """Work around sys.path differences due to execnet using `python -c`."""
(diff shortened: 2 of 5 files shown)


src/xdist/workermanage.py
```diff
@@ -261,7 +262,8 @@ def setup(self):
         remote_module = self.config.hook.pytest_xdist_getremotemodule()
         self.channel = self.gateway.remote_exec(remote_module)
         # change sys.path only for remote workers
-        change_sys_path = not self.gateway.spec.popen
+        # restore sys.path from a frozen copy for local workers
+        change_sys_path = _sys_path if self.gateway.spec.popen else None
```
graingert, line 266: we have to use a frozen copy of sys.path because `--import-mode=prepend` changes the sys.path

testing/test_remote.py
```diff
@@ -292,3 +292,17 @@ def test(get_config_parser, request):
     result = testdir.runpytest_subprocess("-n1")
     assert result.ret == 1
     result.stdout.fnmatch_lines(["*usage: *", "*error: my_usage_error"])
+
+
+def test_remote_sys_path(testdir):
+    """Test/document the behavior due to execnet using `python -c`."""
```
graingert, line 298:
```suggestion
    """Work around sys.path differences due to execnet using `python -c`."""
```

testing/test_remote.py
```diff
@@ -234,7 +234,7 @@ def test():
 
 
 def test_remote_inner_argv(testdir):
-    """Test/document the behavior due to execnet using `python -c`."""
+    """Work around sys.path differences due to execnet using `python -c`."""
```
graingert, line 237:
```suggestion
    """Test/document the behavior due to execnet using `python -c`."""
```

RonnyPfannschmidt (approved):
this one is a nice solution for a gnarly quirk

i would propose to eventually put multiprocessing and/or introcuding a python -m execnet.knownn_worker backend to execnet to make this robust even in the face of remote execution

however thats not something i can touch within the next 2 years and i cant expect anyone else to touch it

RonnyPfannschmidt (approved):

PR #696, opened 2021-08-25
Use setproctitle if available to show state

eqvinox:
setproctitle() changes the process name, particularly in top/ps output
on posix-y systems.  This is very useful to get some transparency on
what's going on, especially if a test starts taking a very long time or
straight up hangs.

This is strictly a "bonus goodie", if setproctitle is not available,
it's not used, and any errors are ignored silently.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,6 +17,7 @@ jobs:
           - "py39-pytestlatest"
           - "py38-pytestmain"
           - "py38-psutil"
+          - "py38-setproctitle"
 
         os: [ubuntu-latest, windows-latest]
         include:
@@ -31,6 +32,8 @@ jobs:
           - tox_env: "py38-pytestmain"
             python: "3.8"
           - tox_env: "py38-psutil"
+            python: "3.8"
+          - tox_env: "py38-setproctitle"
             python: "3.8"
 
     steps:
--- a/README.rst
+++ b/README.rst
@@ -317,6 +317,26 @@ Since version 2.0, the following functions are also available in the ``xdist`` m
         """
 
 
+Identifying workers from the system environment
+-----------------------------------------------
+
+*New in version UNRELEASED TBD FIXME*
+
+If the `setproctitle`_ package is installed, ``pytest-xdist`` will use it to
+update the process title (command line) on its workers to show their current
+state.  The titles used are ``[pytest-xdist running] file.py/node::id`` and
+``[pytest-xdist idle]``, visible in standard tools like ``ps`` and ``top`` on
+Linux, Mac OS X and BSD systems.  For Windows, please follow `setproctitle`_'s
+pointer regarding the Process Explorer tool.
+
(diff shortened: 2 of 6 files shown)


nicoddemus (changes requested):
Hi @eqvinox,

Thanks a lot for the contribution, we appreciate it!

I think we still need a few things before we can merge it:

- [ ] We should configure an `extras` so users can install with `pip install pytest-xdist[setproctitle]`, in case we want to pin compatible versions in the future. Probably just change:
  
  https://github.com/pytest-dev/pytest-xdist/blob/5e0b40cea2b0f392cd681bf5e3086262788f174a/setup.py#L21

  To:

  ```python
  extras_require={"testing": ["filelock"], "psutil": ["psutil>=3.0"], "setproctitle": ["proctitle"]},
  ```

- [ ] Add docs to the README, including the above extra and a quick paragraph about what it does (like you did in the PR description).
- [ ] In `tox.ini`, we need to test with this package, which can be done in its own environment (similar to `[testenv:py38-psutil]`). Also configure CI to run this test environment on Linux.

A test would be good, but I can't think of a way to test this without going through complicated hoops, so I think we can skip that.

While the change is pretty trivial, all this is important to ensure we don't break things by accident later.

Finally, I did take a quick look, but can you double check if the same can be done using `psutil` (which we already have the option to use)?

eqvinox:
@nicoddemus hope to have addressed all your comments.  I couldn't really test the tox item since Debian only provides Python 3.7 and 3.9, installing 3.8 would've been a major hassle, sorry about that :disappointed:.

> Finally, I did take a quick look, but can you double check if the same can be done using `psutil`

`psutil` is a purely read-only interface for process listing, I double checked and it doesn't have this feature.

(Also, as a Linux low-level developer, I can tell you `setproctitle()` [the C function] has barely any relationship to *ps*-like functions.  It works by writing the text to a "magic" location in the running process, it doesn't go through `/proc` or a special syscall or anything like that.)

eqvinox:
> A test would be good, but I can't think of a way to test this without going through complicated hoops, so I think we can skip that.

It might be possible with the combination of `psutil` and `setproctitle`, using the former to check back in on the latter.  But it'd probably end up being super flaky with any number of uncontrollable external influences breaking it (e.g. some Linux security modules constrain process visibility in odd ways, or a prefix on the title might be enforced, or the title might be truncated to a rather short limit, or the process might not see its own title, etc…)

nicoddemus:
> It might be possible with the combination of psutil and setproctitle, using the former to check back in on the latter. But it'd probably end up being super flaky with any number of uncontrollable external influences breaking it

Figured as much, thanks!

nicoddemus (approved):

nicoddemus:
Thanks again @eqvinox!

I will prepare a release soon (possibly today)

nicoddemus:
Sorry for the delay, `2.4.0` has been released with this change. 👍

PR #700, opened 2021-09-01
Fix test_fixture_teardown_failure for pytest master

nicoddemus:
`--debug` now receives an optional file, so the way the test
was calling pytest made it seem like it was passing the python file
file as the debug file.


nicoddemus:
Going ahead and merge this to get CI working again.

PR #708, opened 2021-09-24
Use modern options configurations for hooks

nicoddemus:
Noticed while reviewing https://github.com/pytest-dev/pytest/pull/9118.


RonnyPfannschmidt (approved):
:+1:

PR #723, opened 2021-10-30
Modernize packaging, require pytest 6.2, prepare for pytest 7, other changes

bluetech:
Before we release pytest 7 I figured that pytest-xdist should be ready for it.

Please see the commits for all of the changes. I can break it down to multiple PRs if preferred.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,3 +58,19 @@ setproctitle = setproctitle
 
 [flake8]
 max-line-length = 100
+
+[mypy]
+mypy_path = src
+# TODO: Enable this & fix errors.
+# check_untyped_defs = True
+disallow_any_generics = True
+ignore_missing_imports = True
+no_implicit_optional = True
+show_error_codes = True
+strict_equality = True
+warn_redundant_casts = True
+warn_return_any = True
+warn_unreachable = True
+warn_unused_configs = True
+# TODO: Enable this & fix errors.
+# no_implicit_reexport = True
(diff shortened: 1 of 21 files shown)


setup.cfg
```diff
@@ -1,5 +1,75 @@
...
+
+[options.entry_points]
+pytest11 =
+    xdist = xdist.plugin
+    xdist.looponfail = xdist.looponfail
+
+[options.extras_require]
+testing =
+    filelock
+    pytest>=6.2.0
```
nicoddemus, line 56: I don't think we need to duplicate this do we?

setup.cfg
```diff
@@ -1,5 +1,75 @@
...
+psutil = psutil>=3.0
+setproctitle = setproctitle
+
 [flake8]
 max-line-length = 100
+
+[mypy]
+mypy_path = src
+# check_untyped_defs = True
```
nicoddemus, line 65: I think we should remove the commented options, or leave a comment explaining why they should be left uncommented.

tox.ini
```diff
@@ -30,6 +29,14 @@ deps = pytest
 commands =
   pytest {posargs}
 
+[testenv:linting]
```
nicoddemus, line 32: This was returned for local usage, right?

nicoddemus (approved):
Looks great, thanks a lot for working on this.

Left a few minor comments.

tox.ini
bluetech: Right, I find it convenient to have a tox env for pre-commit as I usually don't have it installed globally.

bluetech:
Thanks for the review @nicoddemus I fixed your comments.

PR #727, opened 2021-11-02
Deprecate `--boxed`

bluetech:
This will allow removing the pytest-forked dependency in the future.

Refs #468.


bluetech:
Had some trouble writing a test for this (`config.issue_config_time_warning`) so I removed my attempt. I checked it manually, hopefully that's enough...

RonnyPfannschmidt (approved):

PR #732, opened 2021-11-28
Fix test_warning_captured_deprecated_in_pytest_6

nicoddemus:
This test started to fail in the 'py38-pytestmain' environment, the cause being `PytestRemovedIn7Warning` being raised by the `conftest.py` file of the test itself. Ignoring it is fine, the purpose of the test is to ensure the hook is not called by `pytest-xdist`.


bluetech (approved):
If I understand correctly, the test will keep working even in pytest 7.1. So LGTM.

PR #733, opened 2021-11-28
Create new dist option 'loadgroup'

baekdohyeop:
### Checklist
- [x] Add change log

### Intro
There are currently several options for distributing tests,
but there is still no suitable option for the following cases:

### Case 1
In this case, it is efficient to divide all tests into different sessions.
```python
@pytest.mark.parametrize('param', [A, B, C, D])
def test_something_heavy(param):
    do_something_heavy_test
```

### Case 2
In this case, it is efficient to run all tests in the same session.
```python
def test_something_light_1(heavy_fixture_cannot_filelock):
    do_something_light_test

def test_something_light_2(heavy_fixture_cannot_filelock):
    do_something_light_test

def test_something_light_3(heavy_fixture_cannot_filelock):
    do_something_light_test
```

### Limit
If you use the loadscope option, all tests in Case 1 are performed in same session,
If the load option is used, all tests of Case 2 may be performed in different sessions.

### Suggestion
Use the following xgroup mark and specify the name through the parameter.
Then, tests with the same name are executed in the same session.
```python
@pytest.mark.xgroup(name="same_session")
def test_something_light_1(heavy_fixture_cannot_filelock):
    do_something_light_test

@pytest.mark.xgroup(name="same_session")
def test_something_light_2(heavy_fixture_cannot_filelock):
    do_something_light_test

@pytest.mark.xgroup(name="same_session")
def test_something_light_3(heavy_fixture_cannot_filelock):
    do_something_light_test
```

### DIFF
--- a/changelog/733.feature.rst
+++ b/changelog/733.feature.rst
@@ -1,1 +1,1 @@
-Create new dist option 'loadgroup'
+New ``--dist=loadgroup`` option, which ensures all tests marked with ``@pytest.mark.xdist_group`` run in the same session/worker. Other tests run distributed as in ``--dist=load``.
--- a/changelog/733.feature.rst
+++ b/changelog/733.feature.rst
@@ -1,0 +1,1 @@
+Create new dist option 'loadgroup'
--- a/src/xdist/plugin.py
+++ b/src/xdist/plugin.py
@@ -98,8 +98,7 @@ def pytest_addoption(parser):
             " the same scope to any available environment.\n\n"
             "loadfile: load balance by sending test grouped by file"
             " to any available environment.\n\n"
-            "loadgroup: load balance by sending any pending test or test group"
-            " to any available enviroment.\n\n"
+            "loadgroup: like load, but sends tests marked with 'xdist_group' to the same worker.\n\n"
             "(default) no: run tests inprocess, don't distribute."
         ),
     )
@@ -207,7 +206,7 @@ def pytest_configure(config):
         config.option.forked = True
 
     config_line = (
-        "xgroup: specify group for tests should run in same session."
+        "xdist_group: specify group for tests should run in same session."
         "in relation to one another. " + "Provided by pytest-xdist."
     )
     config.addinivalue_line("markers", config_line)
--- a/src/xdist/plugin.py
+++ b/src/xdist/plugin.py
@@ -86,7 +86,7 @@ def pytest_addoption(parser):
         "--dist",
         metavar="distmode",
         action="store",
-        choices=["each", "load", "loadscope", "loadfile", "no"],
+        choices=["each", "load", "loadscope", "loadfile", "loadgroup", "no"],
         dest="dist",
         default="no",
         help=(
@@ -98,6 +98,8 @@ def pytest_addoption(parser):
             " the same scope to any available environment.\n\n"
             "loadfile: load balance by sending test grouped by file"
             " to any available environment.\n\n"
+            "loadgroup: load balance by sending any pending test or test group"
+            " to any available enviroment.\n\n"
             "(default) no: run tests inprocess, don't distribute."
         ),
     )
@@ -203,6 +205,12 @@ def pytest_configure(config):
         )
         config.issue_config_time_warning(warning, 2)
         config.option.forked = True
+
+    config_line = (
+        "xgroup: specify group for tests should run in same session."
+        "in relation to one another. " + "Provided by pytest-xdist."
+    )
+    config.addinivalue_line("markers", config_line)
 
 
 @pytest.hookimpl(tryfirst=True)
--- a/src/xdist/remote.py
+++ b/src/xdist/remote.py
@@ -120,15 +120,15 @@ class WorkerInteractor:
         # add the group name to nodeid as suffix if --dist=loadgroup
         if config.getvalue("loadgroup"):
             for item in items:
-                try:
-                    mark = item.get_closest_marker("xgroup")
-                except AttributeError:
-                    mark = item.get_marker("xgroup")
-
-                if mark:
-                    gname = mark.kwargs.get("name")
-                    if gname:
-                        item._nodeid = "{}@{}".format(item.nodeid, gname)
+                mark = item.get_closest_marker("xdist_group")
+                if not mark:
+                    continue
+                gname = (
+                    mark.args[0]
+                    if len(mark.args) > 0
+                    else mark.kwargs.get("name", "default")
+                )
+                item._nodeid = "{}@{}".format(item.nodeid, gname)
 
     @pytest.hookimpl
     def pytest_collection_finish(self, session):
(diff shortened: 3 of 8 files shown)


src/xdist/remote.py
```diff
@@ -116,6 +116,20 @@ def run_one_test(self, torun):
...
         )
 
+    def pytest_collection_modifyitems(self, session, config, items):
+        # add the group name to nodeid as suffix if --dist=loadgroup
+        if config.getvalue("loadgroup"):
+            for item in items:
+                try:
+                    mark = item.get_closest_marker("xgroup")
+                except AttributeError:
+                    mark = item.get_marker("xgroup")
```
nicoddemus, line 126: We only support pytest>=6.2 now, so this can be changed to:

```suggestion
                mark = item.get_closest_marker("xgroup")
```

changelog/733.feature.rst
```diff
@@ -0,0 +1 @@
+Create new dist option 'loadgroup'
```
nicoddemus, line 1:
```suggestion
New ``--dist=loadgroup`` option, which ensures all tests marked with ``@pytest.mark.xgroup`` run in the same session/worker. Other tests run distributed as in ``--dist=load``.
```

src/xdist/plugin.py
```diff
@@ -98,6 +98,8 @@ def pytest_addoption(parser):
             " the same scope to any available environment.\n\n"
             "loadfile: load balance by sending test grouped by file"
             " to any available environment.\n\n"
+            "loadgroup: load balance by sending any pending test or test group"
+            " to any available enviroment.\n\n"
```
nicoddemus, line 102:
```suggestion
            "loadgroup: like load, but sends tests marked with 'xgroup' to the same worker.\n\n"
```

src/xdist/scheduler/loadgroup.py
```diff
@@ -0,0 +1,67 @@
...
+         In the above example, scopes will be::
+
+            example/loadsuite/test/test_beta.py::test_beta0
+            example/loadsuite/test/test_delta.py::Delta1::test_delta0
+            example/loadsuite/epsilon/__init__.py::epsilon.epsilon
+            gname
+            gname
+        """
+        if nodeid.rfind("@") > nodeid.rfind("]"):
+            # check the index of ']' to avoid the case: parametrize mark value has '@'
```
nicoddemus, line 64: 👍

src/xdist/remote.py
```diff
@@ -116,6 +116,20 @@ def run_one_test(self, torun):
             "runtest_protocol_complete", item_index=self.item_index, duration=duration
         )
 
+    def pytest_collection_modifyitems(self, session, config, items):
+        # add the group name to nodeid as suffix if --dist=loadgroup
+        if config.getvalue("loadgroup"):
+            for item in items:
+                try:
+                    mark = item.get_closest_marker("xgroup")
```
nicoddemus, line 124: I think it would be best to use `xdist_group` instead, seems more explicit, what do you think?

src/xdist/remote.py
```diff
@@ -116,6 +116,20 @@ def run_one_test(self, torun):
...
+                except AttributeError:
+                    mark = item.get_marker("xgroup")
+
+                if mark:
+                    gname = mark.kwargs.get("name")
```
nicoddemus, line 129: Probably forcing to use `name` as keyword argument will be a bit surprising, we should also support positional argument (`@pytest.mark.xgroup("mygroup")`), otherwise this will get ignored.

```suggestion
                    gname = mark.args[0] if len(mark.args) > 0 else mark.kwargs.get("name")
```

Also not passing a group name should group to a default group, instead of not grouping at all?

Seems reasonable that all tests marked with `@pytest.mark.xgroup` would be grouped together as one group, and others marked with `@pytest.mark.xgroup("mygroup")` would be grouped in another group. What do you think?

testing/acceptance_test.py
```diff
@@ -1326,6 +1326,115 @@ def test_2():
         assert c1 == c2
 
 
+class TestGroupScope:
+    def test_by_module(self, testdir):
```
nicoddemus, line 1330: Those tests look great! Could you also add another test which uses 2 different groups (`@pytest.mark.xgroup(name="group1")` and `@pytest.mark.xgroup(name="group2")` for example)?

nicoddemus (changes requested):
Thanks a lot @baekdohyeop for following up with this!

Left a few comments, please take a look.

Besides, could you please add a description of the new option to the README? Probably showing some code would be helpful there too.

src/xdist/remote.py
```diff
@@ -236,6 +250,7 @@ def remote_initconfig(option_dict, args):
 
 
 def setup_config(config, basetemp):
+    config.option.loadgroup = True if config.getvalue("dist") == "loadgroup" else False
```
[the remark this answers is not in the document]
baekdohyeop: Thank you. 62e50d0

src/xdist/scheduler/loadgroup.py
```diff
@@ -0,0 +1,67 @@
+from .loadscope import LoadScopeScheduling
+from py.log import Producer
+
+
+class LoadGroupScheduling(LoadScopeScheduling):
+    """Implement load scheduling across nodes, but grouping test only has group mark.
+
+    This distributes the tests collected across all nodes so each test is run
```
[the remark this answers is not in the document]
baekdohyeop: Thank you. 62e50d0

src/xdist/remote.py
baekdohyeop: Thank you. 62e50d0

changelog/733.feature.rst
baekdohyeop: Thank you. 62e50d0

src/xdist/plugin.py
baekdohyeop: Thank you. 62e50d0

src/xdist/remote.py
baekdohyeop: Thank you. 62e50d0

src/xdist/remote.py
baekdohyeop: Thank you. 62e50d0

testing/acceptance_test.py
baekdohyeop: Thank you. 62e50d0

baekdohyeop:
@nicoddemus 
I'm not good at English, so I'd appreciate it if you could improve it.
Thank you.

nicoddemus:
> I'm not good at English, so I'd appreciate it if you could improve it.

Your English is very good, don't worry about it. 😁 

I'm merging as is, and plan to give a quick overhaul of the README in a separate PR. Thanks!

nicoddemus (approved):

baekdohyeop:
> 

Thanks!

PR #751, opened 2022-01-29
Add sphinx docs folder and changelog URLs to metadata

### LINKED ISSUE #749: pytest-xdist rtd documentation site problem
It seems that pytest-xdist documentation rtd website exists but never container docs on it, but searching for xdist docs brings https://readthedocs.org/projects/pytest-xdist amont top results which is confusing.

Apparently @hpk42 @nicoddemus and @RonnyPfannschmidt the the ones listed as admins on RTD. I suspect that removing the project from RTD can sort the problem. As I guess that if nobody bothered to add a sphinx config in so many years it will never happen.

To improve the situation I would also add extra metadata to package to link directly the readme as the Documentation, avoiding further confusions.

nicoddemus:
This reorganizes the README into proper sphinx docs to publish in ReadTheDocs (the text was mostly intact, just moved).

Close #749


nicoddemus:
Preview: https://pytest-xdist--751.org.readthedocs.build/en/751/

ssbarnea (approved):

Issue #331, opened 2018-08-14
How to create individual log files for each worker


nbartos:
I'm trying to figure out how individual log files can be created for each worker, in the same manor that pytest --log-file does when not using xdist.

Since the basic pytest --log-file opens with write, not append (see https://github.com/pytest-dev/pytest/issues/3813), there does not seem to be a way of getting a complete log file for a run using xdist with multiple workers.  However, even if it could be opened in append mode, the individual worker IDs would not be present, and having everything mixed together can be more difficult for troubleshooting.

I'm hoping that there is some way of creating a log file for each worker, but I'm not really sure how to do that, if it's even possible.  I tried setting up a logger in pytest_configure, but worker_id does not seem to be available at that time.

nicoddemus:
Actually you can access the worker id with the `PYTEST_XDIST_WORKER` environment variable, does that help?

nbartos:
Yes that works, thanks!

k3v8ns:
@nbartos   plz  help as you described.

nbartos:
@k3v8ns This is basically what I am doing:

```
def pytest_addoption(parser):
    log_help_text = 'Similar to log_file, but %%w will be replaced with a worker identifier.'
    parser.addini('worker_log_file', help=log_help_text)
    log_group = parser.getgroup('logging')
    log_group.addoption('--worker-log-file', dest='worker_log_file', help=log_help_text)

def pytest_configure(config):
    log_file = config.getoption('worker_log_file')
    if log_file is None:
        log_file = config.getini('worker_log_file')
    log_level_opt = config.getoption('log_level')
    if log_level_opt is None:
        log_level_opt = config.getini('log_level')
    if log_level_opt:
        log_level = getattr(logging, log_level_opt.upper(), None)
        if not log_level:
            raise LogError('Invalid log-level option: {}'.format(log_level_opt))
    else:
        log_level = logging.DEBUG
    if log_file:
        logging.basicConfig(
            format='%(asctime)s %(name)s %(levelname)s %(message)s',
            filename=log_file.replace('%w', os.environ.get('PYTEST_XDIST_WORKER', 'master')),
            level=log_level)
```

Then you just need to set `worker_log_file` or `--worker-log-file` to a path that has `%w` in it, which will be replaced by the worker ID (or "master" for the master process).

bondnotanymore:
@nbartos Thanks a lot for the above code, but still few parts of your solution are unclear to me.
[...]

PR #762, opened 2022-03-08
Add instructions to create one file for each worker

marcosanchotene:
This is to add instructions into the documentation to create one file for each worker, as agreed [here](https://github.com/pytest-dev/pytest-xdist/issues/331#issuecomment-1047019487).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,6 +4,11 @@ repos:
     hooks:
     -   id: black
         args: [--safe, --quiet, --target-version, py35]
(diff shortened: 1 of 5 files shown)


nicoddemus (approved):
Thanks @marcosanchotene!

While at it I did some overall doc improvements, adding blacken-docs and using sphinx autodoc. 👍

marcosanchotene:
You're welcome, @nicoddemus ! I'm glad I could assist somehow. This is also a way to give back for pytest's team work on this great tool.

PR #769, opened 2022-03-25
Add ability to share data between nodes and controller

pehala:
# Why?
Once the nodes are branched from the controller, there is no way to transmit information between them, only through pytest events, which is not always enough (e.g. collecting info from fixtures that are run on nodes).

# How?
I created a simple mechanism that will send back dict of serializable data when the node is being shutdown. This means that you have access to all the shared data from nodes in the session_finish hook on the controller. Example of how it works can be seen in the simple test I wrote.

# Appendix
This is a proof of concept and is in no way finished, the point of this PR is to start the discussion if this feature (or something along those lines) is even wanted and if you agree that yes, I will finish the PR and polish it.


RonnyPfannschmidt (changes requested):
At first glance this looks like a disconnected data dump lacking clarity

Unless there is way more clarity as to what this is supposed to support, we can't accept the new api

pehala:
One of the issues that feature like this would solve is https://github.com/pytest-dev/pytest/issues/7767, where there exists a hack how to use that function but it has to be run on the controller if you need to use this from any fixture, it will not do anything as it is executed on the worker node, which doesn't have any way how to synchronize the data with the master.

Basically, I think that it would be good to have some sort of API that would enable sharing of data between nodes and controller in any way. It doesn't have to be this API (If we agree on how the API should look I will gladly implement it in a better way), this is just PoC that something like it can exist and can work.

PR #813, opened 2022-09-04
Cancel shutdown when crashed worker is restarted

amezin:
Apparently, mark_test_pending() does not work if the test is the last (or only)
one scheduled on the worker. Everything is shutting down before the test has
a chance to run.

1) Add TestCrashItem test to expose the bug
2) Cancel shutdown when crashed worker is restarted


nicoddemus (approved):
Thanks @amezin, great work!

PR #821, opened 2022-10-21
Remove --boxed and pytest-forked dependency

### LINKED ISSUE #468: deprecate and remove the --boxed option

### LINKED ISSUE #543: drop --boxed and deprecate looponfail
As the 2.x release is incoming we should take care of that

Looponfail should eventually be reimplemented in core using subprocess and a hook to register the file system watchers (so pytest plugins could support Realtime Filesystem watching)

nicoddemus:
Fixes #468
Fixes #543

### DIFF
--- a/changelog/468.trivial.rst
+++ b/changelog/468.trivial.rst
@@ -1,0 +1,1 @@
+The ``py`` dependency has been dropped.
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,7 +44,6 @@ python_requires = >=3.6
 install_requires =
     execnet>=1.1
     pytest>=6.2.0
-    pytest-forked
 setup_requires = # left empty, enforce using isolated build system
 
 [options.packages.find]
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ pytest-xdist 2.5.0 (2021-12-10)
 Deprecations and Removals
 -------------------------
 
-- `#468 <https://github.com/pytest-dev/pytest-xdist/issues/468>`_: The ``--boxed`` command line argument is deprecated. Install pytest-forked and use ``--forked`` instead. pytest-xdist 3.0.0 will remove the ``--boxed`` argument and pytest-forked dependency.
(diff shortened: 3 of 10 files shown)


asottile (approved):

nicoddemus:
Not so simple I'm afraid... pytest-xdist's own code still uses some `py` functionality, such as `py.log`, `py.local`, etc.

Will go ahead and try to remove that dependency.

asottile:
looks like the `py.log` can be replaced pretty easily with just `functools.partial(print, file=sys.stderr)` (or `lambda *a, **k: None` in the non-debug case) --  I'll send a patch for that one

nicoddemus:
Yeah I was working on that.

Unfortunately just hit a wall: `pytest_xdist_make_scheduler` receives a `py.log` instance. I guess we will need a round of deprecation for that parameter. 😬

asottile:
> Yeah I was working on that.
> 
> Unfortunately just hit a wall: `pytest_xdist_make_scheduler` receives a `py.log` instance. I guess we will need a round of deprecation for that parameter. 😬

I made a class which looks the same as the `py.log` thing -- and kept the usages the same

nicoddemus:
Cool.

I have some WIP replacing `py.local` with `Path` too, but I won't be able to push it in the next few hours... feel free to open a separate PR replacing `py.log` though.

asottile:
alrighty #822 and #823 should be good to go now I think?

asottile (approved):

asottile (approved):

setup.cfg
```diff
@@ -44,7 +44,7 @@ python_requires = >=3.6
 install_requires =
     execnet>=1.1
     pytest>=6.2.0
-    pytest-forked
+    py
```
nicoddemus, line 47: We still use `py` in pytest-xdist's own code, which is being fixed by #822, #823, and #824.

changelog/468.trivial.rst
```diff
@@ -0,0 +1 @@
+The ``py`` dependency has been dropped.
```
nicoddemus, line 1: AFAICT we no longer depend on `py`. 😁

nicoddemus:
Rebased and ready for final review.

RonnyPfannschmidt (approved):
💥

Issue #835, opened 2022-10-28
The package description adevrtizes a deprecated loop-on-fail feature


hroncok:
The package description explicitly mentions the loop-on-fail feature:

https://github.com/pytest-dev/pytest-xdist/blob/7949c14f67b28c2ceb98c0a76fc74ff2bb5fdb96/setup.cfg#L3

Which is deprecated:

https://github.com/pytest-dev/pytest-xdist/blob/7949c14f67b28c2ceb98c0a76fc74ff2bb5fdb96/CHANGELOG.rst#L17

I think the description needs to be updated. Happy to send a PR if you agree.

nicoddemus:
Definitely, thanks @hroncok, the PR is appreciated. 👍

PR #838, opened 2022-10-28
Remove deprecated stuff from the package description

hroncok:
Fixes https://github.com/pytest-dev/pytest-xdist/issues/835


nicoddemus (approved):
Thanks!

PR #842, opened 2022-11-05
Drop Python 3.6 and add Python 3.11

nicoddemus:
* Drop Python 3.6 support.
* Add Python 3.11 support officially.
* Split CI workflows into testing and deploy.


nicoddemus:
Note: we will need to update "required" builds in branch protection, because the name of the workflows have changed.

RonnyPfannschmidt (approved):

PR #851, opened 2022-11-30
Let users configure dist mode in the configuration file

### LINKED ISSUE #789: Option to set the default distribution algorithm
Hi, I have the following scenario:

- I have some tests that need to be run in the same worker due to the fact they use a shared object (I know it's not very idiomatic for unit tests, but it is really the simpler way to check what I need to check)
- I saw the `--dist loadgroup` option, but unfortunately it is not the default, so I should remember to specify every time the command line option
- I can't add it to the `addopts` setting since it errors if I don't specify either `-n` or `-tx`, and I want to be able to choose to run the tests the standard way too.

What could work for me is a setting to set the default algorithm to use when `--dist` is not specified, so that:
- if I run `pytest` standalone, it doesn't run in parallel
- if I run `pytest -n auto`, it uses the algorithm specified in the setting
- if I run `pytest -n auto --dist loadfile`, it overrides the setting with the `loadfile` algorithm.

nicoddemus:
Fix #789

### DIFF
--- a/changelog/789.feature.rst
+++ b/changelog/789.feature.rst
@@ -1,0 +1,6 @@
+Users can now set a default distribution mode in their configuration file:
+
+.. code-block:: ini
+
+    [pytest]
+    addopts = --dist loadscope
--- a/src/xdist/plugin.py
+++ b/src/xdist/plugin.py
@@ -190,7 +190,19 @@ def pytest_addhooks(pluginmanager):
 
 @pytest.hookimpl(trylast=True)
 def pytest_configure(config):
-    if config.getoption("dist") != "no" and not config.getvalue("collectonly"):
+    config_line = (
+        "xdist_group: specify group for tests should run in same session."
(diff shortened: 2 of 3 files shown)


nicoddemus:
Thanks @RonnyPfannschmidt for the review.

> we ought o consider having a ini option for the default dist mode which is cleaner than addopts (similar to the logging plugin

I was writing my rationale why I think it is not necessary and it turned out to be quite verbose, so I decided instead to turn that into an issue in pytest so we can bring more people to the discussion. 👍  

Please read my rationale there, and if you still feel strongly we should have a dedicated option for it, I will be glad to do it in a follow-up (because another small option would not really be the end of the world). It is also arguable that `--dist` should have been a configuration option in the first place.

I think this is worthwhile to mention anyway as it provides a fix for the original issue and cleans up the code a bit, removing what seemed to be a wart that `--dist` without `-n` would give a cryptic message like that.

PR #857, opened 2022-12-20
Add --maxschedchunk CLI option

### LINKED ISSUE #255: Make load scheduler configurable
I have several projects where the distribution of tests runtime is quite scattered, eg:

- 1000 tests of 10ms
- 100 tests of 1 minute

The current load scheduler comes short in this case, as it often ends up sending a batch of slow tests to the same worker.

As a workaround, I use a forked LoadScheduler that uses a fixed queue size (which I use with the minimum value of 2 -> each worker only has one test in its queue at any time):
```
class FixedLocalQueueLoadScheduling(LoadScheduling):  # no cover
    """
    A fork of pytest-xdist default load scheduler that uses a fixed size for workers local queue size.
    """

    def __init__(self, config, log=None, queue_size=2):
        super().__init__(config, log)
        if queue_size < 2:
            raise ValueError('Queue size must be at least 2')
        self.queue_size = queue_size

    def check_schedule(self, node, duration=0):
        if node.shutting_down:
            return

        if self.pending:
            node_pending = self.node2pending[node]
            if len(node_pending) < self.queue_size:
                num_send = self.queue_size - len(node_pending)
                self._send_tests(node, num_send)
        self.log("num items waiting for node:", len(self.pending))

    def schedule(self):
        assert self.collection_is_completed

        # Initial distribution already happened, reschedule on all nodes
        if self.collection is not None:
            for node in self.nodes:
                self.check_schedule(node)
            return

        # allow nodes to have different collections
        if not self._check_nodes_have_same_collection():
            self.log('**Different tests collected, aborting run**')
            return

        # Collections are identical, create the index of pending items.
        self.collection = list(self.node2collection.values())[0]
        self.pending[:] = range(len(self.collection))
        if not self.collection:
            return

        # Send a batch of tests to run. If we don't have at least two
        # tests per node, we have to send them all so that we can send
        # shutdown signals and get all nodes working.
        initial_batch = min(len(self.pending), self.queue_size * len(self.nodes))

        # distribute tests round-robin up to the batch size
        # (or until we run out)
        nodes = cycle(self.nodes)
        for i in range(initial_batch):
            self._send_tests(next(nodes), 1)

        if not self.pending:
            # initial distribution sent all tests, start node shutdown
            for node in self.nodes:
                node.shutdown()
```

It would be nice to have at least one of these propositions implemented in xdist:

1.  Integrate this scheduler (or an even simpler version where queue_size=2)
2.  Make LoadScheduler configurable, so that users can provide initial_batch_size / items_per_node_min / items_per_node_max
3.  When sending a batch of jobs to a node, shuffle like for the initial batch
4. Maybe improve/reduce a bit the defaults settings for initial_batch_size / items_per_node_min / items_per_node_max

### LINKED ISSUE #855: New `LoadScheduling` batch distribution logic doesn't work for long-running tests
The PR https://github.com/pytest-dev/pytest-xdist/pull/812 that implemented new batch distribution logic (and that is now released in 3.0.2) disregards long-running tests.

I have several modules with long-running tests and many of these tests are now scheduled on single worker, instead of being distributed evenly across multiple workers. As a result, total runtime of my testsuite more than doubled from ~ 2.5 hours to 5+ hours.

Are there any workarounds to bring back the original round-robin scheduling?

cc @amezin

amezin:
Maximum number of tests scheduled in one step.

Setting it to 1 will force pytest to send tests to workers one by one - might be useful for a small number of slow tests.

Larger numbers will allow the scheduler to submit consecutive chunks of tests to workers - allows reusing fixtures.

Unlimited if not set.

Fixes #855
Fixes #255

### DIFF
--- a/changelog/855.feature
+++ b/changelog/855.feature
@@ -1,0 +1,2 @@
+Users can now configure scheduling precision using --maxschedchunk command
+line option.
--- a/changelog/855.feature
+++ b/changelog/855.feature
@@ -1,2 +1,2 @@
-Users can now configure scheduling precision using --maxschedchunk command
+Users can now configure ``load`` scheduling precision using ``--maxschedchunk`` command
 line option.
--- a/src/xdist/plugin.py
+++ b/src/xdist/plugin.py
@@ -153,6 +153,19 @@ def pytest_addoption(parser):
             "on every test run."
         ),
     )
+    group.addoption(
+        "--maxschedchunk",
+        action="store",
+        type=int,
+        help=(
+            "Maximum number of tests scheduled in one step. "
+            "Setting it to 1 will force pytest to send tests to workers one by "
+            "one - might be useful for a small number of slow tests. "
+            "Larger numbers will allow the scheduler to submit consecutive "
+            "chunks of tests to workers - allows reusing fixtures. "
+            "Unlimited if not set."
+        ),
+    )
 
     parser.addini(
         "rsyncdirs",
--- a/src/xdist/plugin.py
+++ b/src/xdist/plugin.py
@@ -158,7 +158,7 @@ def pytest_addoption(parser):
         action="store",
         type=int,
         help=(
-            "Maximum number of tests scheduled in one step. "
+            "Maximum number of tests scheduled in one step for --dist=load. "
             "Setting it to 1 will force pytest to send tests to workers one by "
             "one - might be useful for a small number of slow tests. "
             "Larger numbers will allow the scheduler to submit consecutive "
--- a/src/xdist/scheduler/load.py
+++ b/src/xdist/scheduler/load.py
@@ -272,9 +272,10 @@ class LoadScheduling:
             items_per_node = len(self.collection) // len(self.node2pending)
             # take a fraction of tests for initial distribution
             node_chunksize = min(items_per_node // 4, self.maxschedchunk)
+            node_chunksize = max(node_chunksize, 2)
             # and initialize each node with a chunk of tests
             for node in self.nodes:
-                self._send_tests(node, max(node_chunksize, 2))
+                self._send_tests(node, node_chunksize)
 
         if not self.pending:
             # initial distribution sent all tests, start node shutdown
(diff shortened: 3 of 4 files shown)


changelog/855.feature
```diff
@@ -0,0 +1,2 @@
+Users can now configure scheduling precision using --maxschedchunk command
```
nicoddemus, line 1:
```suggestion
Users can now configure ``load`` scheduling precision using ``--maxschedchunk`` command
```

src/xdist/plugin.py
```diff
@@ -153,6 +153,19 @@ def pytest_addoption(parser):
             "on every test run."
         ),
     )
+    group.addoption(
+        "--maxschedchunk",
+        action="store",
+        type=int,
+        help=(
+            "Maximum number of tests scheduled in one step. "
```
nicoddemus, line 161:
```suggestion
            "Maximum number of tests scheduled in one step for --dist=load. "
```

src/xdist/scheduler/load.py
```diff
@@ -265,10 +271,10 @@ def schedule(self):
             # how many items per node do we have about?
             items_per_node = len(self.collection) // len(self.node2pending)
             # take a fraction of tests for initial distribution
-            node_chunksize = max(items_per_node // 4, 2)
+            node_chunksize = min(items_per_node // 4, self.maxschedchunk)
```
nicoddemus, line 274: For clarity let's keep all the steps to compute `node_chunksize` together:

```suggestion
            node_chunksize = min(items_per_node // 4, self.maxschedchunk)
            node_chunksize = max(node_chunksize, 2)
```

src/xdist/scheduler/load.py
```diff
@@ -265,10 +271,10 @@ def schedule(self):
...
             # and initialize each node with a chunk of tests
             for node in self.nodes:
-                self._send_tests(node, node_chunksize)
+                self._send_tests(node, max(node_chunksize, 2))
```
nicoddemus, line 277:
```suggestion
                self._send_tests(node, node_chunksize)
```

nicoddemus (changes requested):
Thanks @amezin!

Please take a look at my comments. :+1:

nicoddemus (approved):
Thanks @amezin!

PR #865, opened 2023-01-11
Fix some LoadScheduling tests

amezin:
The expected number of nodes didn't match throughout the test code.


nicoddemus (approved):
Thanks!

PR #867, opened 2023-01-12
Document limitations for debugging as well as a workaround (#863)

### LINKED ISSUE #863: Suggest using a remote debugger for debugging in documentation
It would be helpful if the documentation suggested to use a remote debugger (like [remote-pdb](https://github.com/ionelmc/python-remote-pdb)) to circumvent the issue that stdin/stdout can't forwarded.

For example, there could be a short paragraph in "Known limitations" that states that while the absence of forwarding of stdin/stdout means that many debuggers don't work, this approach does.

This would have saved me a couple of hours of trying to get around that limitation just now.

ralokt:
Closes #863

### DIFF
--- a/changelog/863.doc
+++ b/changelog/863.doc
@@ -1,0 +1,1 @@
+Document limitations for debugging due to standard I/O of workers not being forwarded. Also, mention remote debugging as a possible workaround.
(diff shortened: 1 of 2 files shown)


nicoddemus:
Thanks @ralokt, really appreciate it!

ralokt:
Glad to help! :)

PR #870, opened 2023-01-17
Tests: Unset PYTEST_XDIST_AUTO_NUM_WORKERS when the behavior without the envvar is asserted

hroncok:

    ```
    ```

### DIFF
--- a/changelog/870.trivial
+++ b/changelog/870.trivial
@@ -1,0 +1,1 @@
+Make the tests pass even when ``$PYTEST_XDIST_AUTO_NUM_WORKERS`` is set.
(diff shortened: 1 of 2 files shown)


hroncok:
Alternatively, this could be an autouse fixture in the file that does this. Let me know if you prefer that.

RonnyPfannschmidt (approved):
Lovely, thanks, 

as nitpick it might be nice to add `raising=` in front of the False

hroncok:
> as nitpick it might be nice to add `raising=` in front of the False

Done.

nicoddemus (approved):
Thanks!

Issue #402, opened 2019-01-13
logging not capture


zsluedem:
test_5.py file content
```python
import logging
logger = logging.getLogger(__file__)

def test_1():
    logger.info('123123')

def test_2():
    logger.info('123123')

def test_3():
    logger.info('123123')

def test_4():
    logger.info('123123')
```

pytest.ini
```
[pytest]
log_cli=true
log_cli_level=DEBUG
log_format = %(asctime)s %(levelname)s (%(threadName)-10s) %(filename)s:%(lineno)d %(message)s
log_date_format = %Y-%m-%d %H:%M:%S
```

run `pytest test/test_5 -n auto`

logs:
```
================================================================= test session starts =================================================================
platform darwin -- Python 3.6.5, pytest-4.1.1, py-1.7.0, pluggy-0.8.1 -- /Users/w/.virtualenvs/pytest_xdist_fixture_scope_test/bin/python
cachedir: .pytest_cache
rootdir: /Users/w/pytest_xdist_fixture_scope_test, inifile: pytest.ini
plugins: xdist-1.26.0, forked-1.0.1
[gw0] darwin Python 3.6.5 cwd: /Users/w/pytest_xdist_fixture_scope_test
[gw1] darwin Python 3.6.5 cwd: /Users/w/pytest_xdist_fixture_scope_test
[gw2] darwin Python 3.6.5 cwd: /Users/wpytest_xdist_fixture_scope_test
[gw3] darwin Python 3.6.5 cwd: /Users/wpytest_xdist_fixture_scope_test
[gw0] Python 3.6.5 (default, Apr 25 2018, 14:23:58)  -- [GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.1)]
[gw1] Python 3.6.5 (default, Apr 25 2018, 14:23:58)  -- [GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.1)]
[gw2] Python 3.6.5 (default, Apr 25 2018, 14:23:58)  -- [GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.1)]
[gw3] Python 3.6.5 (default, Apr 25 2018, 14:23:58)  -- [GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.1)]
gw0 [4] / gw1 [4] / gw2 [4] / gw3 [4]
scheduling tests via LoadScheduling

[...]

PR #883, opened 2023-03-03
added code to honor --log-cli-level (issue #402)

kurt-cb:
this fixes issue #402 

    ```
    ```

### DIFF
--- a/src/xdist/dsession.py
+++ b/src/xdist/dsession.py
@@ -10,7 +10,8 @@ from xdist.scheduler import (
     LoadGroupScheduling,
     WorkStealingScheduling,
 )
-
+import logging
+import pickle
 
 from queue import Empty, Queue
 
@@ -275,6 +276,17 @@ class DSession:
         rep.node = node
         self.config.hook.pytest_runtest_logreport(report=rep)
         self._handlefailures(rep)
+
+    def worker_runtest_logmessage(self, node, record):
+        record = pickle.loads(record)
+        for handler in logging.getLogger().handlers:
+
+            if not True: #self.respect_handler_level:
+                process = True
+            else:
+                process = record.levelno >= handler.level
+            if process:
+                handler.handle(record)
 
     def worker_runtest_protocol_complete(self, node, item_index, duration):
         """
--- a/src/xdist/remote.py
+++ b/src/xdist/remote.py
@@ -11,6 +11,9 @@ import sys
 import os
 import time
 from typing import Any
+import pickle
+import logging
+import copy
 
 import pytest
 from execnet.gateway_base import dumps, DumpError
@@ -55,6 +58,43 @@ def worker_title(title):
         # changing the process name is very optional, no errors please
         pass
 
+class RemoteMessageHandler(logging.Handler):
+    """
+    This handler sends events to a queue. Typically, it would be used together
+    with a multiprocessing Queue to centralise logging to file in one process
+    (in a multi-process application), so as to avoid file write contention
+    between processes.
+
+    This code is new in Python 3.2, but this class can be copy pasted into
+    user code for use with earlier Python versions.
+    """
+
+    def __init__(self, queue):
+        """
+        Initialise an instance, using the passed queue.
+        """
+        logging.Handler.__init__(self)
+        self.queue = queue
+
+    def emit(self, record):
+        """
+        Emit a record.
+
+        Writes the LogRecord to the queue, preparing it for pickling first.
+        """
+        try:
+            msg = self.format(record)
+            # bpo-35726: make copy of record to avoid affecting other handlers in the chain.
+            record = copy.copy(record)
+            record.message = msg
+            record.msg = msg
+            record.args = None
+            record.exc_info = None
+            record.exc_text = None
+            x = pickle.dumps(record)
+            self.queue.send_log(x)
+        except Exception as e:
+            self.handleError(record)
 
 class WorkerInteractor:
     SHUTDOWN_MARK = object()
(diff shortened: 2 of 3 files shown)


kurt-cb:
This change implements a logger relay handler that pumps the log message back to the master node, which then uses the master logger to display the log events.

To use it, simply set the --log-cli-level option as in normal pytest, and it will push the log level requested.

I believe that is what was asked for in #402

nicoddemus:
Hey @kurt-cb,

Thanks a lot for the contribution, creating a new event to send log events specifically is a good idea. 👍 

Some extra things that would be important to have to get this merged:

* Tests that exercise the feature, preferably under different log settings (to at least avoid regressions).
* A Changelog entry.

Also left a few comments in the code, please take a look. 👍

src/xdist/dsession.py
```diff
@@ -283,6 +284,16 @@ def worker_testreport(self, node, rep):
         self.config.hook.pytest_runtest_logreport(report=rep)
         self._handlefailures(rep)
 
+    def worker_runtest_logmessage(self, node, record):
+        record = pickle.loads(record)
+        for handler in logging.getLogger().handlers:
+            if not True:  # self.respect_handler_level:
```
nicoddemus, line 290: Commented out code?

src/xdist/remote.py
```diff
@@ -56,6 +59,45 @@ def worker_title(title):
...
 
 
+class RemoteMessageHandler(logging.Handler):
+    """
+    This handler sends events to a queue. Typically, it would be used together
+    with a multiprocessing Queue to centralise logging to file in one process
+    (in a multi-process application), so as to avoid file write contention
+    between processes.
+
+    This code is new in Python 3.2, but this class can be copy pasted into
```
nicoddemus, line 69: Seems like this was copied from somewhere, can you add a link to where it was copied from?

src/xdist/remote.py
```diff
@@ -56,6 +59,45 @@ def worker_title(title):
...
+        try:
+            msg = self.format(record)
+            # bpo-35726: make copy of record to avoid affecting other handlers in the chain.
+            record = copy.copy(record)
+            record.message = msg
+            record.msg = msg
+            record.args = None
+            record.exc_info = None
+            record.exc_text = None
+            x = pickle.dumps(record)
```
nicoddemus, line 95: Probably does not matter much, but we might force protocol version 4 here to ensure compatibility between different python interpreters.

src/xdist/remote.py
```diff
@@ -56,6 +59,45 @@ def worker_title(title):
...
+    This code is new in Python 3.2, but this class can be copy pasted into
+    user code for use with earlier Python versions.
+    """
+
+    def __init__(self, queue):
+        """
+        Initialise an instance, using the passed queue.
+        """
+        logging.Handler.__init__(self)
+        self.queue = queue
```
nicoddemus, line 78: A bit confusing calling this `queue` when it is actually a `WorkerInteractor` instance, perhaps use `worker_interactor`? Also worth updating the docs.

src/xdist/remote.py
```diff
@@ -86,6 +136,9 @@ def sendevent(self, name, **kwargs):
         self.log("sending", name, kwargs)
         self.channel.send((name, kwargs))
 
+    def send_log(self, record):
+        self.sendevent("runtest_logmessage", record=record)
```
nicoddemus, line 140: Small suggestion to make things easier to understand:

```suggestion
    def send_log(self, pickled_record):
        self.sendevent("runtest_logmessage", pickled_record=pickled_record)
```

WurmD:
Hi @kurt-cb , shall we pair program to appease @nicoddemus and get this merged? (I'd like to do it for you but don't know 

```
            if not True:  # self.respect_handler_level:
```

```
class RemoteMessageHandler(logging.Handler):
    """
    This handler sends events to a queue. Typically, it would be used together
    with a multiprocessing Queue to centralise logging to file in one process
    (in a multi-process application), so as to avoid file write contention
    between processes.
    This code is new in Python 3.2, but this class can be copy pasted into
```

how to resolve the comments on these two bits

kurt-cb:
I gave up on this PR months ago.  It was raised back in March.  Anyway, the first issue, just need to get rid of the commented out code `# self.respect_handler_le`

The RemoteMessageHandler is based off of the Python QueueHandler that is part of python, here: https://docs.python.org/3/library/logging.handlers.html#queuehandler
Source: https://github.com/python/cpython/blob/8f324b7ecd2df3036fab098c4c8ac185ac07b277/Lib/logging/handlers.py#L1412

kurt-cb:
@WurmD 
I addressed the items you were unsure about.

src/xdist/dsession.py
kurt-cb: removed dead code.

kurt-cb:
@WurmD as for the other issues, I don't have time right now to make those changes and test.  If you would like to do it, go ahead.  You can pull my branch and create a PR to cli_log.  Once you push the first PR, I'll approve you as a contributor.

src/xdist/remote.py
WurmD: `            x = pickle.dumps(record, protocol=4)`
or
`            x = pickle.dumps(record, protocol=pickle.DEFAULT_PROTOCOL)`
?

WurmD:
Here @kurt-cb , would you give me permissions to push to your branch?

Screenshot 2023-09-29 at 12 11 07

src/xdist/remote.py
RonnyPfannschmidt: are log messages pickle safe?
im "pretty" sure there some hazard when dealing with custom objects

WurmD:
> @WurmD as for the other issues, I don't have time right now to make those changes and test. If you would like to do it, go ahead. You can pull my branch and create a PR to cli_log. Once you push the first PR, I'll approve you as a contributor.

Give me permissions to push to your branch please? @kurt-cb

nicoddemus:
@WurmD you can just pull his changes into your own fork, and open a new PR: his original commits will still be correctly attributed.

kurt-cb:
@WurmD, you should be able to make a PR on my repo.  Once I accept the PR I can give you rights to approve.

kurt-cb:
@WurmD you need to fork kurt-cb/pytest-xdist (make sure you get all branches when you do this), then checkout cli_log, make your changes , then create a PR against kurt-cb/pytest-xdist.  Once you do this, I will have the option to make you a contributor.  I don't think I can do anything until you do this.

kurt-cb:
@WurmD I figured out how to make you a collaborator.  You should be able to create wormd/cli_log branch and push your changes directly to there.  then you can create a PR and I will review your changes and push them back here.

WurmD:
K, successfully PRed https://github.com/kurt-cb/pytest-xdist/pull/3

PR #885, opened 2023-03-09
Fix hang caused by `steal` command with empty test queue

### LINKED ISSUE #884: --dist=worksteal + high core counts leads to hangs
I'm able to consistently reproduce with https://github.com/pyca/cryptography on an M1 MBP, which has 10 cores (8 performance, 2 efficiency).

Specifically `py.test -n 10 --dist worksteal --last-failed tests/x509/` consistently hangs.

It consistently works if I remove the `--dist worksteal` flag. Using lower core counts (e.g. `-n 8`) makes the hangs more intermittent, but the hangs still occur.

The hangs happen at the end of the test suite -- I believe after all the tests have run.

If there's any additional information I can provide, pleaese let me know.

amezin:
Fixes #884

### DIFF
--- a/changelog/884.bugfix
+++ b/changelog/884.bugfix
@@ -1,0 +1,1 @@
+Fixed hang in ``worksteal`` scheduler.
--- a/src/xdist/remote.py
+++ b/src/xdist/remote.py
@@ -58,6 +58,7 @@ def worker_title(title):
 
 class WorkerInteractor:
     SHUTDOWN_MARK = object()
+    QUEUE_REPLACED_MARK = object()
 
     def __init__(self, config, channel):
         self.config = config
@@ -71,6 +72,15 @@ class WorkerInteractor:
 
     def _make_queue(self):
         return self.channel.gateway.execmodel.queue.Queue()
+
+    def _get_next_item_index(self):
+        """Gets the next item from test queue. Handles the case when the queue
+        is replaced concurrently in another thread.
+        """
+        result = self.torun.get()
+        while result is self.QUEUE_REPLACED_MARK:
+            result = self.torun.get()
+        return result
 
     def sendevent(self, name, **kwargs):
         self.log("sending", name, kwargs)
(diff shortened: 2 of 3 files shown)


changelog/884.bugfix
```diff
@@ -0,0 +1 @@
+Fix hang caused by `steal` command with empty test queue (`worksteal` scheduler).
```
nicoddemus, line 1:
```suggestion
Fixed hang using ``worksteal`` scheduler using a high number of CPU cores.
```

src/xdist/remote.py
```diff
@@ -72,6 +73,12 @@ def __init__(self, config, channel):
     def _make_queue(self):
         return self.channel.gateway.execmodel.queue.Queue()
 
+    def _get_next_item_index(self):
```
nicoddemus, line 76: It would be nice to add a docstring explaining why the retry is needed.

nicoddemus (approved):
Left a few minor suggestions, otherwise LGTM. :+1: 

Thanks for tackling this so quickly.

After this gets merged, feel free to prepare a new release.

changelog/884.bugfix
nicoddemus: The rationale here is that changelogs are read by end-users which are not usually familiarized with the internal, so better use a language closer to the usage of the plugin.

src/xdist/remote.py
amezin: Will `RETRY_MARK` -> `QUEUE_CHANGED_MARK` make the code more self-explanatory?

changelog/884.bugfix
amezin: But "high number of CPU cores" has nothing to do with the issue. I can reproduce it with `-n 3`

nicoddemus: Ahh OK, feel free to reword it. :+1:

src/xdist/remote.py
nicoddemus: Ahh I think so. I would still add a docstring describing the issue however.

reaperhulk:
Verified this stops the hanging we were seeing in the `pyca/cryptography` test suite. Thanks @amezin!

nicoddemus (approved):

PR #898, opened 2023-05-02
Fix year in project documentation

n8creator:
Fixed an outdated year in the project documentation

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,7 +18,7 @@
 # -- Project information -----------------------------------------------------
 
 project = "pytest-xdist"
-copyright = "2022, holger krekel and contributors"
+copyright = "2023, holger krekel and contributors"
 author = "holger krekel and contributors"
 
 master_doc = "index"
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt:
This needs to be the start year or a range, current year is always wrong

RonnyPfannschmidt (changes requested):
I missed when this was made wrong the first time, we need start year or a range

nicoddemus:
I suggest just a start date, we can use the one for the first tag:

image

We do that in pytest: https://github.com/pytest-dev/pytest/blob/271bdf6c235360131ae5a3e3426f39174811c0cf/LICENSE#LL3C15-L3C20

cclauss:
See the highest-ranked answer...
https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

nicoddemus:
Updated it to use the copyright for the first commit. 👍

nicoddemus:
Thanks @n8creator!

cclauss (approved):

PR #901, opened 2023-05-10
Make collection progress output less verbose

### LINKED ISSUE #555: Revamp progress output when collecting workers
We might want to review the progress message that happens while workers are being collected, because some people find it too verbose (https://github.com/pytest-dev/pytest-xdist/issues/423), or wonder what `gw` stands for (that's gateway, which an execnet concept).

Currently we have:

## default

collecting:

```
==============================
plugins: forked-1.2.0, xdist-1.32.1
gw0 [0] / gw1 [0]
```

done:

```
==============================
plugins: forked-1.2.0, xdist-1.32.1
gw0 [156] / gw1 [156]
```

## verbose

collecting:

```
==============================
plugins: xdist-1.31.0
[gw0] win32 Python 3.7.6 cwd: d:\projects\pytest-xdist
[gw1] win32 Python 3.7.6 cwd: d:\projects\pytest-xdist
[gw0] Python 3.7.6 (tags/v3.7.6:43364a7ae0, Dec 19 2019, 00:42:30) [MSC v.1916 64 bit (AMD64)]
[gw1] Python 3.7.6 (tags/v3.7.6:43364a7ae0, Dec 19 2019, 00:42:30) [MSC v.1916 64 bit (AMD64)]
gw0 [0] / gw1 [0]
```

done:

```
==============================
plugins: xdist-1.31.0
[gw0] win32 Python 3.7.6 cwd: d:\projects\pytest-xdist
[gw1] win32 Python 3.7.6 cwd: d:\projects\pytest-xdist
[gw0] Python 3.7.6 (tags/v3.7.6:43364a7ae0, Dec 19 2019, 00:42:30) [MSC v.1916 64 bit (AMD64)]
[gw1] Python 3.7.6 (tags/v3.7.6:43364a7ae0, Dec 19 2019, 00:42:30) [MSC v.1916 64 bit (AMD64)]
gw0 [156] / gw1 [156]
```

---

To remove the `gw` term and reduce the output problems when having too many workers, how about:

## default

collecting:

```
==============================
plugins: forked-1.2.0, xdist-1.32.1
collecting: 1/8 workers...
```

done:

```
==============================
plugins: forked-1.2.0, xdist-1.32.1
2 workers [156 items]
```

## verbose

collecting:

```
==============================
plugins: xdist-1.31.0
[w0] win32 Python 3.7.6 cwd: d:\projects\pytest-xdist
[w1] win32 Python 3.7.6 cwd: d:\projects\pytest-xdist
[w0] Python 3.7.6 (tags/v3.7.6:43364a7ae0, Dec 19 2019, 00:42:30) [MSC v.1916 64 bit (AMD64)]
[w1] Python 3.7.6 (tags/v3.7.6:43364a7ae0, Dec 19 2019, 00:42:30) [MSC v.1916 64 bit (AMD64)]
collecting: 1/8 workers...
```

done:

```
==============================
plugins: xdist-1.31.0
[w0] win32 Python 3.7.6 cwd: d:\projects\pytest-xdist
[w1] win32 Python 3.7.6 cwd: d:\projects\pytest-xdist
[w0] Python 3.7.6 (tags/v3.7.6:43364a7ae0, Dec 19 2019, 00:42:30) [MSC v.1916 64 bit (AMD64)]
[w1] Python 3.7.6 (tags/v3.7.6:43364a7ae0, Dec 19 2019, 00:42:30) [MSC v.1916 64 bit (AMD64)]
2 workers [156 items]
```

And in case we are running workers for the same interpreter, we even don't print the `[w*]` lines with interpreter/xdist information because they are redundant.

nicoddemus:
Fix #555

### DIFF
--- a/src/xdist/dsession.py
+++ b/src/xdist/dsession.py
@@ -1,3 +1,8 @@
+from __future__ import annotations
+import sys
+from enum import Enum, auto
+from typing import Sequence
+
 import pytest
 
 from xdist.remote import Producer
@@ -251,14 +256,16 @@ class DSession:
         self._session.testscollected = len(ids)
         self.sched.add_node_collection(node, ids)
         if self.terminal:
-            self.trdist.setstatus(node.gateway.spec, "[%d]" % (len(ids)))
+            self.trdist.setstatus(
+                node.gateway.spec, WorkerStatus.CollectionDone, tests_collected=len(ids)
+            )
         if self.sched.collection_is_completed:
             if self.terminal and not self.sched.has_pending:
                 self.trdist.ensure_show_status()
                 self.terminal.write_line("")
                 if self.config.option.verbose > 0:
                     self.terminal.write_line(
-                        "scheduling tests via %s" % (self.sched.__class__.__name__)
+                        f"scheduling tests via {self.sched.__class__.__name__}"
                     )
             self.sched.schedule()
 
@@ -345,7 +352,7 @@ class DSession:
         if rep.failed:
             self.countfailures += 1
             if self.maxfail and self.countfailures >= self.maxfail:
-                self.shouldstop = "stopping after %d failures" % (self.countfailures)
+                self.shouldstop = f"stopping after {self.countfailures} failures"
 
     def triggershutdown(self):
         self.log("triggering shutdown")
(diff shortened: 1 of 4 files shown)


src/xdist/dsession.py
```diff
@@ -457,3 +485,34 @@ def get_default_max_worker_restart(config):
...
+    if status_and_items and all(s == WorkerStatus.CollectionDone for s in statuses):
+        first = status_and_items[0]
+        status, tests_collected = first
+        tests_noun = "item" if tests_collected == 1 else "items"
+        return f"{total_workers} {workers_noun} [{tests_collected} {tests_noun}]"
+    if WorkerStatus.CollectionDone in statuses:
+        done = sum(1 for s, c in status_and_items if c > 0)
+        return f"collecting: {done}/{total_workers} {workers_noun}"
+    if WorkerStatus.Ready in statuses:
+        ready = list(statuses).count(WorkerStatus.Ready)
```
bluetech, line 509: statuses is already a list, so the `list()` is redundant (also below).

src/xdist/dsession.py
```diff
@@ -457,3 +485,34 @@ def get_default_max_worker_restart(config):
...
+    status_and_items: Sequence[tuple[WorkerStatus, int]]
+) -> str:
+    """
+    Return the line to display during worker setup/collection based on the
+    status of the workers and number of tests collected for each.
+    """
+    statuses = [s for s, c in status_and_items]
+    total_workers = len(statuses)
+    workers_noun = "worker" if total_workers == 1 else "workers"
+    if status_and_items and all(s == WorkerStatus.CollectionDone for s in statuses):
```
bluetech, line 500: is `not status_and_items` possible?

src/xdist/dsession.py
```diff
@@ -457,3 +485,34 @@ def get_default_max_worker_restart(config):
...
+        first = status_and_items[0]
```
bluetech, line 501: I didn't understand at first why this only looks at the first worker. Would add a comment to clarify: "All workers collect the same items.". Also maybe write it as `tests_collected = status_and_items[0][1]`. Also maybe `tests_collected` -> `items_collected`, to match the terminology?

src/xdist/dsession.py
```diff
@@ -372,32 +379,51 @@ def handle_crashitem(self, nodeid, worker):
...
+    """Status of each worker during creation/collection."""
+
+    # Worker spec has just been created.
+    Created = auto()
+
+    # Worker has been initialized.
+    Initialized = auto()
+
+    # Worker is now ready for collection.
+    Ready = auto()
```
bluetech, line 392: If you can bare the length, would name it `ReadyForCollection`, because `Ready` is somewhat ambiguous (I would have thought that the "collection done" one is "ready").

bluetech (approved):
:+1:  from me

src/xdist/dsession.py
nicoddemus: In the current design not really, but given the type cannot prevent that, I thought better guard against that, because `all([])` returns True and will break on the next line.

src/xdist/dsession.py
nicoddemus: Done!

src/xdist/dsession.py
nicoddemus: Done, thanks!

src/xdist/dsession.py
nicoddemus: Thanks for catching that, this code was refactored and I did not remember to change this.

RonnyPfannschmidt (approved):
This is a little gnarly, but the ux is no longer 🎊

adamchainz:
Thank you ❤️

PR #905, opened 2023-05-12
Release 3.3.0

nicoddemus:
We will also use this opportunity to test the new deploy workflow. 👍 

(Note that now release branches *must* be pushed to the main repository, no longer to the forks).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,6 +10,11 @@ on:
     branches:
       - "*"
 
+# Cancel running jobs for the same workflow and branch.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
 
   check-package:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,6 +10,14 @@ on:
       - "*"
 
 jobs:
+
+  check-package:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Build and Check Package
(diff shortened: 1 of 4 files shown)


bluetech (approved):

RonnyPfannschmidt (approved):

.github/workflows/test.yml
```diff
@@ -3,13 +3,27 @@ name: test
...
       - "*"
 
+# Cancel running jobs for the same workflow and branch.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
+
+  check-package:
```
nicoddemus, line 20: It would be nice to produce the package once, using this job, and then reuse this package for all the 'test' job, but I couldn't find an easy way to ask for `tox` to use a specific wheel to test (instead of building its own).

nicoddemus: Actually I just discovered the `--installpkg` from `tox run` which does exactly that.

Will try it out in a separate PR.

nicoddemus:
2 hickups:

1. It published successfully, but we failed to push the tag because of the addition of `--annotate`:

   ```
   Committer identity unknown
   
   *** Please tell me who you are.
   
   Run
   
     git config --global user.email "you@example.com"
     git config --global user.name "Your Name"
   
   to set your account's default identity.
   Omit --global to set the identity only in this repository.
   ```

2. The 'deploy' environment did not ask for approval because I missed that configuration in the [Settings](https://github.com/pytest-dev/pytest-xdist/settings/environments/1034918748/edit), but I have fixed it now.



I pushed the tag manually for now, will open a separate PR attempting to fix 1).

PR #938, opened 2023-08-14
Fix incompatibility error with `--pdb` despite using `-n0`

### LINKED ISSUE #937: Plugin raises error with `--pdb` option despite using `-n0`
```toml
# pyproject.toml
[tool.pytest.ini_options]
addopts = "--dist=load"
```

```commandline
$ pytest  --pdb -n0
ERROR: --pdb is incompatible with distributing tests; try using -n0 or -nauto.
```

As you can see,  `-n0` is already passed. However, pytest still suggests using `-n0` to address the issue.

#### Solution

`-n0` should imply `--dist=no`.

atugushev:
Fixes #937.

### DIFF
--- a/testing/test_plugin.py
+++ b/testing/test_plugin.py
@@ -94,17 +94,12 @@ def test_auto_detect_cpus(
     check_options(config)
     assert config.getoption("numprocesses") == 99
 
-    config = pytester.parseconfigure("-nauto", "--pdb")
-    check_options(config)
-    assert config.getoption("usepdb")
-    assert config.getoption("numprocesses") == 0
-    assert config.getoption("dist") == "no"
-
-    config = pytester.parseconfigure("-nlogical", "--pdb")
-    check_options(config)
-    assert config.getoption("usepdb")
-    assert config.getoption("numprocesses") == 0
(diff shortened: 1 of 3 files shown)


nicoddemus:
Rebased.

GitHub does not always run the GH actions for first time contributors, IIRC there were people opening random PRs which would use the actions for mining... 👀

testing/test_plugin.py
```diff
@@ -48,6 +48,23 @@ def test_dist_options(pytester: pytest.Pytester) -> None:
...
+    check_options(config)
+    assert config.option.dist == "no"
+    assert config.option.tx == []
+
+    config = pytester.parseconfigure(
+        "--numprocesses", "0", "--dist", "each", "--tx", "2*popen"
+    )
+    check_options(config)
+    assert config.option.dist == "no"
+    assert config.option.tx == []
```
nicoddemus, line 66: Don't we want to also test passing `-n0` and `--pdb` explicitly?

atugushev: Surely we do, thanks! Addressed in 0de6180.

nicoddemus (approved):
Thanks!

PR #963, opened 2023-11-05
Properly wait for workers when test run terminates early.

paul-ollis:
Currently, a reason to terminate early (e.g. test failure with --exitfail
option set) causes DSession to immediately raise an Interrupt exception.
Subsequent reports generated by the workers, during the shutdown phase, are
discarded.

One consequence is that, for the failing test, teardown and testfinish are
ignored, which prevents corresponding hooks pytest_runtest_logreport and
pytest_runtest_logfinish being executed for the failing test (this problem
covered in #54). The reporting of tests executing in other workers is also left in an
indeterminate state. This can affect other plugin code.

This is a relatively simple fix, which appears to have minimal and, I think,
acceptable impact on text execution behaviour. 

The observable differences are differences in what is reported about a test
run. For example, when running, for example with the '-x/--exitfail' option, it
is possible that  more than a single test failure is reported. This is because
more than one test did fail before the test run was completely stopped. The
reporting is absolutely correct; and complete. Prior to this change, only a
single failure would have been reported, but because of incomplete and arguably
incorrect reporting.

### DIFF
--- a/src/xdist/dsession.py
+++ b/src/xdist/dsession.py
@@ -122,8 +122,7 @@ class DSession:
         while not self.session_finished:
             self.loop_once()
             if self.shouldstop:
-                if not self.shuttingdown:
-                    self.triggershutdown()
+                self.triggershutdown()
                 pending_exception = Interrupted(str(self.shouldstop))
         if pending_exception:
             raise pending_exception
@@ -363,10 +362,11 @@ class DSession:
                 self.shouldstop = f"stopping after {self.countfailures} failures"
 
     def triggershutdown(self):
-        self.log("triggering shutdown")
-        self.shuttingdown = True
-        for node in self.sched.nodes:
-            node.shutdown()
+        if not self.shuttingdown:
+            self.log("triggering shutdown")
+            self.shuttingdown = True
+            for node in self.sched.nodes:
+                node.shutdown()
 
     def handle_crashitem(self, nodeid, worker):
         # XXX get more reporting info by recording pytest_runtest_logstart?
--- a/src/xdist/dsession.py
+++ b/src/xdist/dsession.py
@@ -118,11 +118,14 @@ class DSession:
         assert self.sched is not None
 
         self.shouldstop = False
+        pending_exception = None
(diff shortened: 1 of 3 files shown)


src/xdist/dsession.py
```diff
@@ -118,11 +118,15 @@ def pytest_runtestloop(self):
         assert self.sched is not None
 
         self.shouldstop = False
+        pending_exception = None
         while not self.session_finished:
             self.loop_once()
             if self.shouldstop:
-                self.triggershutdown()
-                raise Interrupted(str(self.shouldstop))
+                if not self.shuttingdown:
```
nicoddemus, line 125: Perhaps it would be safer/less error prone to move this `if not self.shuttingdown` to `triggershutdown()`, so it becomes a no-op if `self.shuttingdown` is already True?

nicoddemus (approved):
Thanks @paul-ollis for the contribution!

Left a small comment, otherwise the change looks good.

Could you also add a CHANGELOG entry please? I think `improvement` would be adequate here.

src/xdist/dsession.py
paul-ollis: I like that. I'll make the change.

paul-ollis: Bit puzzled about the change log entry. Is 'improvement' supposed to be the $type in the name of the changelog file? I ask because it is not one of the defined types, if my understanding is correct,

nicoddemus: Yeah, something like `changelog/963.improvement.rst` . 👍

paul-ollis: OK. Done.

nicoddemus:
I will leave this open for a few days in case others want to give this a review, otherwise I will merge then.

RonnyPfannschmidt (approved):
its a nice enhancement based on the current codebase and a remidner that we really ought to have some more testable statemachines as actually unittestable entities

nicoddemus:
Released in `3.4.0`, thanks @paul-ollis!

PR #1020, opened 2024-02-12
pyproject.toml: Move metadata and setuptools config here from setup.cfg

mkoeppe:
Prepared using ini2toml.

    ```
    ```


RonnyPfannschmidt (commented):
Thanks, looking good

Please add a change log segment

RonnyPfannschmidt (approved):

nicoddemus (approved):
Thanks @mkoeppe!

Moved flake8 and mypy configuration as well, so we can drop `setup.cfg` entirely.

PR #1026, opened 2024-02-23
Terminate the test loop if `shouldfail` or `shouldstop` is set

### LINKED ISSUE #1024: Session-scoped fixture executed too many times with `-x` / `--maxfails`
This is related, but I think slightly different than #868 and #420 (which deals with the fact that these flags do not stop execution after a fail).

Observe this test file:
```python
import sys
from time import sleep

from pytest import fixture

# hack for live prints (https://stackoverflow.com/questions/27006884)
sys.stdout = sys.stderr


@fixture(scope='session', autouse=True)
def session_fixture():
    print('### executing fixture! ###')


def test_1():
    assert False


def test_2():
    sleep(1)


def test_3():
    sleep(1)


def test_4():
    sleep(1)
```

Running this with: 
```
pytest test_parallel.py  -n 2  -s --maxfail=1
```

leads to the `session_fixture` being executed 3 times!   Removing `--maxfail=1`, you can see `session_fixture` is executed twice (as expected).

bbrown1867:
Terminate the test loop if `shouldfail` or `shouldstop` is set in a worker's pytest `session`

Closes #1024.

### DIFF
--- a/src/xdist/dsession.py
+++ b/src/xdist/dsession.py
@@ -182,18 +182,17 @@ class DSession:
             self.shouldstop = f"{node} received keyboard-interrupt"
             self.worker_errordown(node, "keyboard-interrupt")
             return
-
         shouldfail = node.workeroutput["shouldfail"]
         shouldstop = node.workeroutput["shouldstop"]
         for shouldx in [shouldfail, shouldstop]:
             if shouldx:
-                self.shouldstop = shouldx
-                self._active_nodes.remove(node)
-                return
-
-        if node in self.sched.nodes:
-            crashitem = self.sched.remove_node(node)
-            assert not crashitem, (crashitem, node)
+                if not self.shouldstop:
+                    self.shouldstop = shouldx
+                break
+        else:
+            if node in self.sched.nodes:
+                crashitem = self.sched.remove_node(node)
+                assert not crashitem, (crashitem, node)
         self._active_nodes.remove(node)
 
     def worker_internal_error(self, node, formatted_error):
--- a/src/xdist/dsession.py
+++ b/src/xdist/dsession.py
@@ -182,7 +182,7 @@ class DSession:
             self.shouldstop = f"{node} received keyboard-interrupt"
             self.worker_errordown(node, "keyboard-interrupt")
             return
-        if not self.shouldstop and node in self.sched.nodes:
+        if node in self.sched.nodes:
             crashitem = self.sched.remove_node(node)
             assert not crashitem, (crashitem, node)
         self._active_nodes.remove(node)
--- a/src/xdist/dsession.py
+++ b/src/xdist/dsession.py
@@ -182,7 +182,16 @@ class DSession:
             self.shouldstop = f"{node} received keyboard-interrupt"
             self.worker_errordown(node, "keyboard-interrupt")
             return
-        if not self.shouldstop and node in self.sched.nodes:
+
+        shouldfail = node.workeroutput["shouldfail"]
+        shouldstop = node.workeroutput["shouldstop"]
+        for shouldx in [shouldfail, shouldstop]:
+            if shouldx:
+                self.shouldstop = shouldx
+                self._active_nodes.remove(node)
+                return
+
+        if node in self.sched.nodes:
             crashitem = self.sched.remove_node(node)
             assert not crashitem, (crashitem, node)
         self._active_nodes.remove(node)
--- a/src/xdist/dsession.py
+++ b/src/xdist/dsession.py
@@ -182,7 +182,7 @@ class DSession:
             self.shouldstop = f"{node} received keyboard-interrupt"
             self.worker_errordown(node, "keyboard-interrupt")
             return
-        if node in self.sched.nodes:
+        if not self.shouldstop and node in self.sched.nodes:
             crashitem = self.sched.remove_node(node)
             assert not crashitem, (crashitem, node)
         self._active_nodes.remove(node)
--- a/src/xdist/remote.py
+++ b/src/xdist/remote.py
@@ -155,6 +155,8 @@ class WorkerInteractor:
         self.nextitem_index = self._get_next_item_index()
         while self.nextitem_index is not self.SHUTDOWN_MARK:
             self.run_one_test()
+            if session.shouldfail or session.shouldstop:
+                break
         return True
 
     def run_one_test(self):
--- a/src/xdist/remote.py
+++ b/src/xdist/remote.py
@@ -154,12 +154,10 @@ class WorkerInteractor:
         self.channel.setcallback(self.handle_command, endmarker=self.SHUTDOWN_MARK)
         self.nextitem_index = self._get_next_item_index()
         while self.nextitem_index is not self.SHUTDOWN_MARK:
-            self.run_one_test()
-            if session.shouldfail or session.shouldstop:
-                break
+            self.run_one_test(session.shouldfail or session.shouldstop)
         return True
 
-    def run_one_test(self):
+    def run_one_test(self, skip: bool):
         self.item_index = self.nextitem_index
         self.nextitem_index = self._get_next_item_index()
 
@@ -173,7 +171,10 @@ class WorkerInteractor:
         worker_title("[pytest-xdist running] %s" % item.nodeid)
 
         start = time.time()
-        self.config.hook.pytest_runtest_protocol(item=item, nextitem=nextitem)
+
+        if not skip:
+            self.config.hook.pytest_runtest_protocol(item=item, nextitem=nextitem)
+
         duration = time.time() - start
 
         worker_title("[pytest-xdist idle]")
--- a/src/xdist/remote.py
+++ b/src/xdist/remote.py
@@ -154,10 +154,12 @@ class WorkerInteractor:
         self.channel.setcallback(self.handle_command, endmarker=self.SHUTDOWN_MARK)
         self.nextitem_index = self._get_next_item_index()
         while self.nextitem_index is not self.SHUTDOWN_MARK:
-            self.run_one_test(session.shouldfail or session.shouldstop)
+            self.run_one_test()
+            if session.shouldfail or session.shouldstop:
+                break
         return True
 
-    def run_one_test(self, skip: bool):
+    def run_one_test(self):
         self.item_index = self.nextitem_index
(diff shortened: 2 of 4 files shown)


src/xdist/remote.py
```diff
@@ -155,6 +155,8 @@ def pytest_runtestloop(self, session):
         self.nextitem_index = self._get_next_item_index()
         while self.nextitem_index is not self.SHUTDOWN_MARK:
             self.run_one_test()
+            if session.shouldfail or session.shouldstop:
+                break
```
bbrown1867, line 159: There might be a better way to do this - perhaps testing `self.torun` for _any_ `self.SHUTDOWN_MARK` items, even if those items aren't the "next" ones in the queue? I can't find much API info on what data type `torun` actually is though.

src/xdist/dsession.py
```diff
@@ -182,7 +182,7 @@ def worker_workerfinished(self, node):
             self.shouldstop = f"{node} received keyboard-interrupt"
             self.worker_errordown(node, "keyboard-interrupt")
             return
-        if node in self.sched.nodes:
+        if not self.shouldstop and node in self.sched.nodes:
```
bbrown1867, line 185: Without this gating, the following error would be occur using the example code in #1024:

```
INTERNALERROR> AssertionError: ('test_parallel.py::test_2', <WorkerController gw0>)
```

src/xdist/remote.py
nicoddemus: This change seems reasonable enough to me. 👍

testing/acceptance_test.py
```diff
@@ -109,12 +109,12 @@ def test_skip():
         )
         assert result.ret == 1
 
-    def test_exitfail_waits_for_workers_to_finish(
+    def test_exitfirst_waits_for_workers_to_finish(
```
bbrown1867, line 112: `-x` is short for `--exitfirst`, not `--exitfail`. With the change in this PR, there are no passing tests - each worker hits the "global" limit failure limit of 1 locally, and exit.

src/xdist/dsession.py
bluetech: I assume pytest-xdist interprets the worker existing early as it crashing? Is gw0 the worker that reached the local maxfail, or the other one?

testing/acceptance_test.py
```diff
@@ -138,15 +138,16 @@ def test_fail6():
                 time.sleep(0.3)
         """
         )
+        # Two workers are used
         result = pytester.runpytest(p1, "-x", "-rA", "-v", "-n2")
         assert result.ret == 2
+        # DSession should stop when the first failure is reached. Two failures
+        # may actually occur, due to timing.
         result.stdout.re_match_lines([".*Interrupted: stopping.*[12].*"])
-        m = re.search(r"== (\d+) failed, (\d+) passed in ", str(result.stdout))
```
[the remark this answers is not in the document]
bluetech: Actually it will be better to use `result.parse_outcomes()` (I don't remember exactly, I think it's called that)

src/xdist/dsession.py
bbrown1867: Yeah that's correct, `gw0` is the worker that is reaching the local maxfail in the example shown in my comment.

testing/acceptance_test.py
```diff
@@ -1180,6 +1181,23 @@ def test_aaa1(crasher):
...
+        """
+        def test1():
+            assert False
+
+        def test2():
+            pass
+    """
+    )
+    result = pytester.runpytest_subprocess("--maxfail=1", "-n 1")
+    result.stdout.re_match_lines([".* 1 failed in .*"])
```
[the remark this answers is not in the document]
bbrown1867: Fixed in a022a1cec601b936dfd03cac67757bd9f0fa9acd

testing/acceptance_test.py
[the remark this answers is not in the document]
bbrown1867: Fixed in a022a1cec601b936dfd03cac67757bd9f0fa9acd

bbrown1867:
@bluetech are we good to merge?

src/xdist/dsession.py
bluetech: Can you explain what is going on here then -- why is `node` still in `self.sched.nodes` even though it exited cleanly and not crashed?

bluetech: Just a reminder, I never really worked on the pytest-xdist code so I'm mostly trying to interpret things just like you :)

> My interpretation of self.sched.nodes is that it is a fixed-list, it doesn't get updated/reduced when workers stop running.

Right.

>  If the worker completed sucessfully, crashitem would be None.

This is true.

> In our case it did not complete succesfully - it aborted early - so crashitem is not None.

I think it *should* complete successfully, where I define "complete successfully" as "did not crash" (as opposed to "had any test failures" or such).

What is the worker <-> coordinator communication that needs to happen for the `shouldstop` scenario to not be considered a crash?

bbrown1867: > Just a reminder, I never really worked on the pytest-xdist code so I'm mostly trying to interpret things just like you :)

Understood! Sorry to keep tagging you, just not sure who else to get feedback from. Does this project have active maintainers?

> What is the worker <-> coordinator communication that needs to happen for the shouldstop scenario to not be considered a crash?

Sadly I don't really know. I see that, in `remote.py`, the worker calls `self.sendevent` in a number of scenarios - for example, `runtest_protocol_complete`, testreport`, `internal_error`, etc. I tried inserting some of these `sendevent` calls into the `pytest_runtestloop` to see if that leads to more graceful shutdown. Unforatunely nothing seems to work. I think I'm lacking sufficient knowledge to go any further.

bbrown1867:
@bluetech  I found a new solution that does the same thing (stopping the worker when the global max is reached) without needing any changes to `dsession.py`. Let me know what you think!

bluetech:
If I understand the new solution, when reaching `shouldstop` it "drains" the remaining items from the worker by fake-executing them and reporting them to the DSession. I have a feeling that this faking is going to cause some issues since we're basically skipping over the entire interior of the runtest protocol. Perhaps the assumption is that by this point, the DSession has already processed its own `shouldstop` and is preparing to shutdown, but this is the sort of race we want to avoid.

Reading the code, I'm thinking perhaps we go back to something like the previous solution, but make it safe.
Previously the worker would just stop, and the dsession would interpret it as a crash, which it might as well have been, no way to know.
I think instead we can add the worker's `shouldstop` to the `worker_workerfinished` message. If the dsession sees it, it knows it's not a crash and it can set its own `shouldstop` (if not already set) and shutdown.

I'm not sure it would work out, but if does I think it will fix the issue cleanly. WDYT?

bbrown1867:
> I'm not sure it would work out, but if does I think it will fix the issue cleanly. WDYT?

Yeah I agree that is a cleaner solution and I like that it propagates the worker session `shouldstop`/`shouldfail` status to the `DSession`. I took a stab at this in c927b1a3af5f040552861d03a1dac8a69940ca46 and it seems to work on the example from #1024.

```
(.venv) % pytest test_parallel.py  -n 2  -s --maxfail=1
======================================================================== test session starts =========================================================================
platform darwin -- Python 3.8.13, pytest-8.0.0, pluggy-1.4.0
configfile: tox.ini
plugins: xdist-0.1.dev1152+ge11ad4d
2 workers [4 items]
### executing fixture! ###
### executing fixture! ###
F..
============================================================================== FAILURES ==============================================================================
_______________________________________________________________________________ test_1 _______________________________________________________________________________
[gw0] darwin -- Python 3.8.13 /Users/bbrown/Development/fun/pytest-xdist/.venv/bin/python

    def test_1():
>       assert False
E       assert False

test_parallel.py:16: AssertionError
====================================================================== short test summary info =======================================================================
FAILED test_parallel.py::test_1 - assert False
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! xdist.dsession.Interrupted: stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
==================================================================== 1 failed, 2 passed in 2.49s =====================================================================
```

The failure occurs on `gw0` and it exits. `gw1` runs both tests successfully, finishing before the failure propagates from `gw0` to `DSession`.

src/xdist/dsession.py
```diff
@@ -182,6 +182,15 @@ def worker_workerfinished(self, node):
             self.shouldstop = f"{node} received keyboard-interrupt"
             self.worker_errordown(node, "keyboard-interrupt")
             return
+
+        shouldfail = node.workeroutput["shouldfail"]
+        shouldstop = node.workeroutput["shouldstop"]
+        for shouldx in [shouldfail, shouldstop]:
+            if shouldx:
+                self.shouldstop = shouldx
```
bluetech, line 190: If `shouldstop` is already set (which I think is possible if raced with another worker?), I think we shouldn't overwrite it.

src/xdist/dsession.py
```diff
@@ -182,6 +182,15 @@ def worker_workerfinished(self, node):
...
+                self._active_nodes.remove(node)
```
bluetech, line 191: Let's write the code such that this line is not repeated below. Can use `for-break-else` or whatever way you like.

bluetech (changes requested):
Thanks for the update, I think the approach is good now. I left a couple of comments on the code.

bluetech (approved):
Thanks! I will merge in a few days, maybe @RonnyPfannschmidt or @nicoddemus want to take a look as well.

bluetech:
@bbrown1867 Can you add a `changelog/1024.bugfix` changelog entry? Would do it myself but no permission to push to your fork. I wrote this entry if you want to use it:

```rst
Added proper handling of ``shouldstop`` (such as set by ``--max-fail``) and ``shouldfail`` conditions in workers.
Previously, a worker might have continued executing further tests before the controller could terminate the session.
```

PR #1027, opened 2024-02-25
Use execnet main_thread_only execmodel

zmedico:
Use the execnet `main_thread_only `execmodel so that code which expects to run in the main thread will just work.  This execmodel has been merged to the execnet master branch via pytest-dev/execnet#243, so this patch should not be merged until there is a released version of execnet supporting the `main_thread_only` execmodel.

Also increase minimum python version to 3.8 since execnet dropped 3.7 support in pytest-dev/execnet#245.

Closes: pytest-dev/pytest-xdist#620

### DIFF
--- a/changelog/620.bugfix
+++ b/changelog/620.bugfix
@@ -1,0 +1,1 @@
+Use the ``execnet`` new ``main_thread_only`` "execmodel" so that code which expects to only run in the main thread will now work as expected.
--- a/changelog/1027.improvement.rst
+++ b/changelog/1027.improvement.rst
@@ -1,0 +1,3 @@
+``pytest-xdist`` workers now always execute the tests in the main thread.
+
+Previously some tests might end up executing in a separate thread other than ``main`` in the workers, due to some internal `èxecnet`` details. This can cause problems specially with async frameworks where the event loop is running in the ``main`` thread (for example `#620 <https://github.com/pytest-dev/pytest-xdist/issues/620>`__).
--- a/src/xdist/looponfail.py
+++ b/src/xdist/looponfail.py
@@ -82,7 +82,7 @@ class RemoteControl:
             print("RemoteControl:", msg)
 
     def initgateway(self) -> execnet.Gateway:
-        return execnet.makegateway("popen")
+        return execnet.makegateway("execmodel=main_thread_only//popen")
 
     def setup(self) -> None:
         if hasattr(self, "gateway"):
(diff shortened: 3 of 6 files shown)


RonnyPfannschmidt (approved):
As gnarly as required by the hack, thanks

pyproject.toml
```diff
@@ -25,16 +25,15 @@ classifiers = [
...
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
 ]
-requires-python = ">=3.7"
+requires-python = ">=3.8"
 dependencies = [
-    "execnet>=1.1",
+    "execnet@git+https://github.com/pytest-dev/execnet#egg=master",
```
nicoddemus, line 36: Once a new execnet is released:

```suggestion
    "execnet>=2.1.0",
```

changelog/620.bugfix
```diff
@@ -0,0 +1,3 @@
+Use the execnet main_thread_only execmodel so that code which expects to run in the main thread will just work.
+
+Also increase minimum python version to 3.8 since execnet dropped 3.7 support in pytest-dev/execnet#245.
```
nicoddemus, line 3:
```suggestion
Use the ``execnet`` new ``main_thread_only`` "execmodel" so that code which expects to only run in the main thread will now work as expected.
```

Let's mention the Python 3.7 drop into its own note. :+1:

nicoddemus (changes requested):
Great work, thanks @zmedico!

Leaving this as "Request changes" to make it clear this is waiting on a new execnet release. :+1:

bluetech:
@zmedico we now depend on execnet>=2.1.0 in master so I think this unblocked. @zmedico can you rebase? Note that master already dropped py3.7 as well so you can remove these changes from the PR changelog notes.

bluetech:
@zmedico The just-merged typing PR created other conflicts (sorry about that, I forgot). Can you rebase again? (if you want I can do it as well).

@nicoddemus You marked "requested changes" but now that we depend on new execnet I assume you're good with merging?

After this is merged I can do the 3.6.0 release.

nicoddemus (approved):
Thanks @zmedico!

nicoddemus:
> After this is merged I can do the 3.6.0 release.

That would be great, thanks @bluetech!

webknjaz:
xref https://github.com/pytest-dev/pytest-xdist/pull/1070#issuecomment-2067290245

PR #1047, opened 2024-03-23
Remove pytest InvocationParams backward compat

bluetech:
It's available since pytest 5.2, and pytest-xdist requires pytest>=6.2.


nicoddemus (approved):
Good find!

PR #1060, opened 2024-04-05
Pre-typing fixes/improvements

bluetech:
These are some commits I made while adding typing to the plugin (will send a PR after this one is merged).


nicoddemus (approved):
Reviewed each commit individually, looks great, thanks!

PR #1061, opened 2024-04-06
Add a Scheduling Protocol

bluetech:
- Clearly delineate the required interface of schedulers
- Useful for typing

### DIFF
--- a/src/xdist/dsession.py
+++ b/src/xdist/dsession.py
@@ -15,6 +15,7 @@ from xdist.scheduler import LoadFileScheduling
 from xdist.scheduler import LoadGroupScheduling
 from xdist.scheduler import LoadScheduling
 from xdist.scheduler import LoadScopeScheduling
+from xdist.scheduler import Scheduling
 from xdist.scheduler import WorkStealingScheduling
 from xdist.workermanage import NodeManager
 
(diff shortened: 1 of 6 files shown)


src/xdist/dsession.py
```diff
@@ -97,17 +98,21 @@ def pytest_collection(self):
         return True
 
     @pytest.hookimpl(trylast=True)
-    def pytest_xdist_make_scheduler(self, config, log):
+    def pytest_xdist_make_scheduler(self, config, log) -> Scheduling | None:
         dist = config.getvalue("dist")
-        schedulers = {
```
bluetech, line 102: Not using the dict because I don't want to include the `__init__` in the protocol (if that's even possible...)

src/xdist/dsession.py
```diff
@@ -97,17 +98,21 @@ def pytest_collection(self):
...
+            return LoadScheduling(config, log)
+        if dist == "loadscope":
+            return LoadScopeScheduling(config, log)
+        if dist == "loadfile":
+            return LoadFileScheduling(config, log)
+        if dist == "loadgroup":
+            return LoadGroupScheduling(config, log)
+        if dist == "worksteal":
+            return WorkStealingScheduling(config, log)
+        return None
```
bluetech, line 115: There is a semantic change here -- if `dist` is not one of the builtin schedulers, this hook now returns `None` instead of raising a KeyError before. This seems better to me, allowing other hookimpls (e.g. wrappers) to handle this more gracefully.

src/xdist/scheduler/each.py
```diff
@@ -103,6 +103,9 @@ def mark_test_complete(self, node, item_index, duration=0):
     def mark_test_pending(self, item):
         raise NotImplementedError()
 
+    def remove_pending_tests_from_node(self, node, indices):
```
bluetech, line 106: Some schedulers were missing this method that is used by `DSession`, if guess it's not called for them so raising `NotImplementedError`.

src/xdist/scheduler/protocol.py
```diff
@@ -0,0 +1,47 @@
+from __future__ import annotations
+
+from typing import Protocol
+from typing import Sequence
+
+from xdist.workermanage import WorkerController
+
+
+class Scheduling(Protocol):
```
nicoddemus, line 9: Minor, up to you: I think a more appropriate name would be `Scheduler` (a subject), "sheduling" is a verb AFAIK. But I understand also that all the implementations use `Scheduling` in their name, so on one hand it makes sense to name this `Scheduling`, on the other hand, "two wrongs don't make a right".

Up to you really. 👍

nicoddemus (approved):
I assume we will not add a changelog for this change because this is not public yet, correct?

src/xdist/scheduler/protocol.py
bluetech: Yes, I don't know why the existing schedulers are called `Scheduling` but I don't think we can safely change them at this point. And then I think being consistent is better.

PR #1065, opened 2024-04-15
Improve typing

### LINKED ISSUE #1057: Add strict/complete typing
The current typing is very imcomplete. To make the plugin more maintainable and understandable, add typing to all of the currently untyped code and apply stricter mypy config.

Prerequisite tasks:
- execnet typing (done, needs an execnet release + bump execnet requirement)
- bump min pytest requirement from 6.2 to 7.0 -- while 6.2 is typed it is missing a lot of stuff, 7.0 is the first version with mostly complete typing

bluetech:
This adds type annotations such that we pass `mypy --strict`. The typings are not exposed as that would need some work on the public API.

Fix #1057.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,3 +32,4 @@ repos:
           - pytest>=7.0.0
           - execnet>=2.1.0
           - types-psutil
+          - setproctitle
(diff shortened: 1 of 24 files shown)


nicoddemus (approved):
Looks great, thanks!

PR #1072, opened 2024-04-20
Cache execnet gateway rinfo during WorkerController setup

zmedico:
Cache execnet gateway info during WorkerController setup for backward
compatibility, in order to avoid a later main_thread_only deadlock
error triggered when pytest-cov calls rinfo after the main thread is
already busy. See pytest-dev/execnet#274 for corresponding test case.

Fixes: 20e3ac774e8f ("Use execnet main_thread_only execmodel (#1027)")

I was able not able to reproduce the issue using a small project with pytest-xdist and pytest-cov.

I was able to reproduce it by running tox in a checkout of setuptools, after using pip to install pytest-xdist-3.6.0 into the .tox/py virtualenv. Then I applied this PR to the virtualenv and that fixed the issue, with pytest-cov successfully producing coverage results.

### DIFF
--- a/src/xdist/workermanage.py
+++ b/src/xdist/workermanage.py
@@ -312,6 +312,11 @@ class WorkerController:
 
     def setup(self) -> None:
         self.log("setting up worker session")
+        # Cache rinfo for backward compatibility, since pytest-cov
+        # accesses rinfo while the main thread is busy executing our
+        # remote_exec call, which triggers a deadlock error for the
+        # main_thread_only execmodel if the rinfo has not been cached.
+        self.gateway._rinfo()
         spec = self.gateway.spec
         args = [str(x) for x in self.config.invocation_params.args or ()]
         option_dict = {}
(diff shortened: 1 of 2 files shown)


bluetech:
This change LGTM, pending https://github.com/pytest-dev/execnet/pull/274#issuecomment-2080397459 (and if we decide to fix in xdist instead of execnet, should update the "Remove this after execnet is fixed" comment)

src/xdist/workermanage.py
```diff
@@ -312,6 +312,11 @@ def shutting_down(self) -> bool:
 
     def setup(self) -> None:
         self.log("setting up worker session")
+        # Cache rinfo for backward compatibility, since pytest-cov
+        # accesses rinfo while the main thread is busy executing our
+        # remote_exec call, which triggers a deadlock error for the
+        # main_thread_only execmodel if the rinfo has not been cached.
+        self.gateway._rinfo()
```
zmedico, line 319: > This change LGTM, pending [pytest-dev/execnet#274 (comment)](https://github.com/pytest-dev/execnet/pull/274#issuecomment-2080397459) (and if we decide to fix in xdist instead of execnet, should update the "Remove this after execnet is fixed" comment)

Okay, comment updated.

bluetech (approved):
Thanks!

If there are no further comments, I'll merge this and release 3.6.1.

nicoddemus (approved):

PR #1090, opened 2024-05-31
Add empty sentinel item in WorkerInteractor queue

yuvalino:
When pytest_runtest_protocol is called with item==nextitem multiple times, with a test that accepts some fixtures -> the second call forward raises an error.

This is causes in WorkerInteractor when the same test appears twice in a row in its items queue.

To deal with this, we define an "empty" item that shall allow us to call pytest_runtest_protocol with nextitem=None, effectively solving this problem so long sequential executions of the same items are delimited by these empty items in the queue.


yuvalino:
Much better solution has been proposed in #1091

PR #1092, opened 2024-06-03
Clarify error message to indicate that users should refer to the documentation of pytest-xdist

namurphy:
This PR updates an error message to better indicate where users should look for more information (i.e., the pytest-xdist docs rather than the pytest docs).

## Motivation

When I encountered this error message recently, there wasn't any nearby context that indicated where the "Known limitations" section was.  It took me a minute to figure out that it wasn't from pytest itself but rather from pytest-xdist.  

## Checklist

    ```
    ```

Also: thank you to everyone who created and contributed to this fantastic extension! 🥇

### DIFF
--- a/changelog/1092.trivial
+++ b/changelog/1092.trivial
@@ -1,0 +1,1 @@
+Update an error message to better indicate where users should go for more information.


nicoddemus (approved):
Thanks, appreciate it!

PR #1118, opened 2024-08-21
Make node ID modification more stable.

criemen:
We force test collection on the controller node, as we run some global setup there before starting the workers. That setup also depends on the test node IDs, but currently, pytest-xdist only modifies the test node IDs on the workers, that've been started and where `setup_config` has been called. I don't see a good reason for this, and for creating a separate config entry, so let's just check the right condition in the pytest hook. That way, the hook will always do the same node ID transformation, regardless of pytest-xdist setup.

    ```
    ```

### DIFF
--- a/src/xdist/remote.py
+++ b/src/xdist/remote.py
@@ -207,7 +207,7 @@ class WorkerInteractor:
         items: list[pytest.Item],
     ) -> None:
         # add the group name to nodeid as suffix if --dist=loadgroup
-        if config.getvalue("loadgroup"):
+        if config.getoption("dist") == "loadgroup":
             for item in items:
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt:
Node id modification needs to be dropped all together

criemen:
That is a good point - as all the unit tests fail on this PR anyways, I'll see what I can do.

RonnyPfannschmidt:
Doing the stuff the hack with node id enabled in a more natural way is a major undertaking

Only got there as a follow-up

criemen:
I'll collect feedback on https://github.com/pytest-dev/pytest-xdist/pull/1119 and if that's not a viable route, I'll try to resurrect this PR instead.

PR #1125, opened 2024-09-03
Publish package with attestations

nicoddemus:
Follow up to https://github.com/pytest-dev/pytest-xdist/pull/1123.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -40,6 +40,8 @@ jobs:
 
     - name: Publish package to PyPI
       uses: pypa/gh-action-pypi-publish@v1.10.0
+      with:
+        attestations: true
 
     - name: Push tag
       run: |


webknjaz (approved):

RonnyPfannschmidt (approved):
Do we actually need the checkout action, we might be able to do the deploy without, but that's a low hanging research topic

nicoddemus:
> Do we actually need the checkout action, we might be able to do the deploy without, but that's a low hanging research topic

A quick look indicates we need to be checked out in order to push the tag.

.github/workflows/deploy.yml
```diff
@@ -40,6 +40,8 @@ jobs:
 
     - name: Publish package to PyPI
       uses: pypa/gh-action-pypi-publish@v1.10.0
```
webknjaz, line 42: @nicoddemus so there is a bug in v1.10.0, and it explodes on attestation code path. It's fixed in v1.10.1 FYI.

Issue #1014, opened 2024-01-30
A new distribution mode is needed to group test cases, execute Group 1 first, and then Group 2


zhupengfarmer:
I have many test cases that need to wait for all other cases to be executed before executing. However, the current grouping method does not support it

ciaconet:
Hi, we are also hoping for this same feature, are there any plans to add something like this? For more context: we have many test cases that consume different amounts of RAM/CPU. We'd like to separate these cases into groups based on pytest marks, then have high resource consuming tests run in an isolated group w/some number of workers (for example, 4 workers), then run tests with lower resource consumption in a separate group with 8 workers.

PR #1130, opened 2024-09-18
Create Scheduler for Grouping Tests w/ Variable number of Workers

ciaconet:
This pull request addresses [Issue 1014](https://github.com/pytest-dev/pytest-xdist/issues/1014). It adds a new scheduler that allows a user to add tests to arbitrary groups using the pytest mark `xdist_custom`. Using the mark a group name and number of workers to use for that group is specified. Groups are ran sequentially but tests in each group are ran in parallel with the given number of workers.

**This pull request is a rough prototype. We aren't too familiar with pytest-xdist internals and we don't expect it is merge-able as is. However, we're hoping to open the conversation to see if there are improvements we can make to get this merged into pytest-xdist.**

Since we wanted groups of tests to execute sequentially, we did not want to schedule tests from a pending group until after the currently running group completed all tests. We found that in order to obtain the status of the final test scheduled on a worker, we needed to send a shutdown signal to the worker. Because of this, between groups a shutdown signal is sent to each worker, and then workers are initialized again prior to scheduling the next group of tests.

We left the directory `xdist-testing-ntop` in as an example but will remove that and add proper tests.

Will add a `changelog/1014.feature` if this PR gains traction.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -39,7 +39,7 @@ jobs:
         path: dist
 
     - name: Publish package to PyPI
-      uses: pypa/gh-action-pypi-publish@v1.10.1
+      uses: pypa/gh-action-pypi-publish@v1.10.3
       with:
         attestations: true
 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -19,7 +19,7 @@ jobs:
     - uses: actions/checkout@v4
 
     - name: Build and Check Package
-      uses: hynek/build-and-inspect-python-package@v2.8
+      uses: hynek/build-and-inspect-python-package@v2.9
 
   deploy:
     needs: package
@@ -39,7 +39,9 @@ jobs:
         path: dist
 
     - name: Publish package to PyPI
-      uses: pypa/gh-action-pypi-publish@v1.9.0
+      uses: pypa/gh-action-pypi-publish@v1.10.1
+      with:
+        attestations: true
 
     - name: Push tag
       run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,6 +41,7 @@ jobs:
           - "py311-pytestlatest"
           - "py311-pytestmain"
           - "py312-pytestlatest"
+          - "py313-pytestlatest"
           - "py310-psutil"
           - "py310-setproctitle"
 
@@ -60,6 +61,8 @@ jobs:
             python: "3.11"
           - tox_env: "py312-pytestlatest"
             python: "3.12"
+          - tox_env: "py313-pytestlatest"
+            python: "3.13"
           - tox_env: "py310-psutil"
             python: "3.10"
           - tox_env: "py310-setproctitle"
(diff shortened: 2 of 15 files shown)


dmakhno:
@ciaconet , I wonder does it address #385 too?
For my need a concept of [pytest-mp](https://github.com/ansible/pytest-mp) (which was archieved) is more suitabe. And this scheduler seems a good fit - and open new strategies for scheduler.

Just my naive visualization on need:
image

And your CR, "custom marker", is exactly "bucket", "batch".

ciaconet:
> @ciaconet , I wonder does it address #385 too? For my need a concept of [pytest-mp](https://github.com/ansible/pytest-mp) (which was archieved) is more suitabe. And this scheduler seems a good fit - and open new strategies for scheduler.
> 
> Just my naive visualization on need: image
> 
> And your CR, "custom marker", is exactly "bucket", "batch".



> @ciaconet , I wonder does it address #385 too? For my need a concept of [pytest-mp](https://github.com/ansible/pytest-mp) (which was archieved) is more suitabe. And this scheduler seems a good fit - and open new strategies for scheduler.
> 
> Just my naive visualization on need: image
> 
> And your CR, "custom marker", is exactly "bucket", "batch".

Unfortunately as is this PR doesn't address https://github.com/pytest-dev/pytest-xdist/issues/385. The custom mark would give you the option to run some tests sequentially, but this would happen in it's own "bucket". With the changes on this PR we can't have some tests run sequentially while simultaneously running other tests in parallel.

vitaly-krugl:
Regarding

> between groups a shutdown signal is sent to each worker, and then workers are initialized again prior to scheduling the next group of tests

That may slow things down considerably due to test collection when workers start up. My workplace has about 30 thousand test cases, and each test collection cycle takes multiple minutes in our use case. 

Each xdist worker performs test collection and filtering independently when it starts up, combined with any additional processing that the test suite might perform via pytest hooks. 

So, if you shut down and restart all xdist workers between each group, you would inject non-trivial test collection/processing delays between the execution of each group. If someone has many groups, they would incur many such delays.

Please correct me if I am wrong.

Heroico:
> Regarding
> 
> > between groups a shutdown signal is sent to each worker, and then workers are initialized again prior to scheduling the next group of tests
> 
> This may slow down the test run considerably due to test collection overhead when workers start up. My workplace has about 30 thousand test cases at this time - and growing fast - and each test collection cycle takes multiple minutes in our use case.
> 
> Each xdist worker performs test collection and filtering independently when it starts up, combined with any additional processing that the test suite might perform via pytest hooks within the scope of each worker.
> 
> So, if you shut down and restart all xdist workers between each group, you would inject non-trivial test collection/processing delays between the execution of each group. If someone has many groups, they would incur many such delays.
> 
> Please correct me if I am wrong.

Shutting down workers and restarting when changing groups indeeds incurs in collection/processing again in this implementation.

In our case, we have around 20000 test cases, and our collection takes about 10 seconds whenever we switch from one group to the next.

Heroico:
We closed this due to lack of activity. Do let us know if the `pytest-xdist` project is interested in a scheduler like this one.

PR #1144, opened 2024-10-23
Make 'steal' command atomic

amezin:
Either unschedule all requested tests, or none if it's not possible - if some of the requested tests have already been processed by the time the request arrives. It may happen if the worker runs tests faster than the controller receives and processes status updates. But in this case maybe it's just better to let the worker keep running.

This is a prerequisite for group/scope support in worksteal scheduler - so they won't be broken up incorrectly.

This change could break schedulers that use "steal" command. However:

1) worksteal scheduler doesn't need any adjustments.

2) I'm not aware of any external schedulers relying on this command yet.

So I think it's better to keep the protocol simple, not complicate it for imaginary compatibility with some unknown and likely non-existent schedulers.

### DIFF
--- a/changelog/1144.feature
+++ b/changelog/1144.feature
@@ -1,0 +1,3 @@
+The internal `steal` command is now atomic - it unschedules either all requested tests or none.
+
+This is a prerequisite for group/scope support in the `worksteal` scheduler, so test groups won't be broken up incorrectly.
--- a/src/xdist/remote.py
+++ b/src/xdist/remote.py
@@ -8,6 +8,7 @@ needs not to be installed in remote environments.
 
 from __future__ import annotations
 
+import collections
 import contextlib
 import enum
 import os
@@ -15,9 +16,11 @@ import sys
 import time
 from typing import Any
 from typing import Generator
+from typing import Iterable
 from typing import Literal
 from typing import Sequence
 from typing import TypedDict
+from typing import Union
 import warnings
 
 from _pytest.config import _prepareconfig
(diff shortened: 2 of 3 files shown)


src/xdist/remote.py
```diff
@@ -146,30 +174,25 @@ def handle_command(
             self.steal(kwargs["indices"])
 
     def steal(self, indices: Sequence[int]) -> None:
```
nicoddemus, line 176: Can we add a docstring here? I think it is important to mention that it will steal all `indices` given if it *can* steal all tests and they are all still pending, or none at all.

src/xdist/remote.py
```diff
@@ -66,7 +69,44 @@ def worker_title(title: str) -> None:
 
 class Marker(enum.Enum):
     SHUTDOWN = 0
-    QUEUE_REPLACED = 1
+
+
+class TestQueue:
+    """A simple queue that can be inspected and modified while the lock is held."""
```
nicoddemus, line 75:
```suggestion
    """A simple queue that can be inspected and modified while the lock is held via the ``lock()`` method."""
```

testing/test_remote.py
```diff
@@ -267,6 +267,10 @@ def test_func4(): pass
 
         worker.sendcommand("steal", indices=[1, 2])
         ev = worker.popevent("unscheduled")
+        assert ev.kwargs["indices"] == []
```
nicoddemus, line 270:
```suggestion
        # Cannot steal index 1 because it is completed already, so do not steal any. 
        assert ev.kwargs["indices"] == []
```

testing/test_remote.py
```diff
@@ -267,6 +267,10 @@ def test_func4(): pass
...
+
+        worker.sendcommand("steal", indices=[2])
```
nicoddemus, line 272:
```suggestion
        # Index 2 can be stolen, as it is still pending.
        worker.sendcommand("steal", indices=[2])
```

changelog/1144.feature
```diff
@@ -0,0 +1 @@
+Make "steal" command atomic - make it unschedule either all requested tests or none.
```
nicoddemus, line 1:
```suggestion
The internal `steal` command is now atomic - make it unschedule either all requested tests or none.

This is a prerequisite for group/scope support in the `worksteal` scheduler, so tests won't be broken up incorrectly.
```

nicoddemus (changes requested):
Thanks @amezin, overall looks good!

I left a few minor comments.

nicoddemus (approved):
LGTM, thanks!

@RonnyPfannschmidt would appreciate a second set of eyes before we merge this.

Issue #1215, opened 2025-06-12
Can no longer group test cases by non-str attribute


sshane:
We used to be able to group test cases in xdist by a hashable class attribute (added by parameterized_class), not it assumes that it's always a string:

https://github.com/commaai/openpilot/blame/fcebb5eb9f8735b6516a70f630c28f297c252c6a/selfdrive/car/tests/test_models.py#L474

`test_route` is a dataclass
```python
@parameterized_class(('platform', 'test_route'), get_test_cases())
@pytest.mark.xdist_group_class_property('test_route')
class TestCarModel(TestCarModelBase):
  pass
```

```
batman@workstation-shane:~/openpilot/selfdrive/car/tests$ pytest -n1 test_models.py 
================================================================================== test session starts ==================================================================================
platform linux -- Python 3.12.4, pytest-8.4.0, pluggy-1.6.0
Using --randomly-seed=3682492429
rootdir: /home/batman/openpilot
configfile: pyproject.toml
plugins: repeat-0.9.4, subtests-0.14.1, timeout-2.4.0, mock-3.14.0, asyncio-0.26.0, randomly-3.16.0, cov-6.1.1, cpp-2.6.0, hypothesis-6.47.5, xdist-3.7.0
asyncio: mode=Mode.STRICT, asyncio_default_fixture_loop_scope=function, asyncio_default_test_loop_scope=function
1 worker [1932 items]  
INTERNALERROR> def worker_internal_error(
INTERNALERROR>         self, node: WorkerController, formatted_error: str
INTERNALERROR>     ) -> None:
INTERNALERROR>         """
INTERNALERROR>         pytest_internalerror() was called on the worker.
INTERNALERROR>     
INTERNALERROR>         pytest_internalerror() arguments are an excinfo and an excrepr, which can't
INTERNALERROR>         be serialized, so we go with a poor man's solution of raising an exception
INTERNALERROR>         here ourselves using the formatted message.
INTERNALERROR>         """
INTERNALERROR>         self._active_nodes.remove(node)
INTERNALERROR>         try:
INTERNALERROR> >           assert False, formatted_error
INTERNALERROR> E           AssertionError: Traceback (most recent call last):
INTERNALERROR> E               File "/home/batman/openpilot/.venv/lib/python3.12/site-packages/_pytest/main.py", line 289, in wrap_session
INTERNALERROR> E                 session.exitstatus = doit(config, session) or 0
[...]

PR #1216, opened 2025-06-12
Ensure all xdist group names are strings

sshane:
Will come back and fill out the description. This is a possible solve of https://github.com/pytest-dev/pytest-xdist/issues/1215 for review.

The previous code before https://github.com/pytest-dev/pytest-xdist/pull/1201 called `__str__` on all group names via the f-string coercion


nicoddemus:
Thanks!

PR #1229, opened 2025-07-26
Parallel node setup

sshane:
When running pytest with a high number of workers `pytest -n32`, I noticed a significant amount of time is just spent on initializing the nodes (2s here), since each call to `self.group.makegateway(spec)` takes ~0.05s. When using a pool, the total time is instead roughly the time it takes to create one Python instance (0.05s).

https://github.com/pytest-dev/pytest-xdist/blob/239def9be9a489526f0a56e6ba442149f17cb919/src/xdist/workermanage.py#L106


sshane:
@nicoddemus I know you reviewed my last PR. There are a fair amount of tests that assert workers pick up tests in a specific order. Is rewriting the tests to not care about order the correct fix, or should we order the nodes to pick up tests deterministically again?

PR #1301, opened 2026-01-31
Support `-X cpu_count` and PYTHON_CPU_COUNT (#1296)

### LINKED ISSUE #1296: Should `pytest_xdist_auto_num_workers` respect `-X cpu_count`/`PYTHON_CPU_COUNT` on Python 3.13+?
#### Background

Python 3.13 introduced new functionality to address the case "how many (logical) CPU cores can this Python program use" via the [`os.process_cpu_count` function][PROCESS_CPU_COUNT]. In particular, it also introduced the related [`-X cpu_count` option][X_OPTION] and [`PYTHON_CPU_COUNT` environment variable][PYTHON_CPU_COUNT], which can be used to override the auto-detected value, thus serving a similar purpose to the `PYTEST_XDIST_AUTO_NUM_WORKERS` environment variable. These two override mechanisms were retrofitted to the [`os.cpu_count` function][CPU_COUNT] as well (the variant that does not take CPU affinity into account).

[PROCESS_CPU_COUNT]: https://docs.python.org/3.13/library/os.html#os.process_cpu_count
[X_OPTION]: https://docs.python.org/3.13/using/cmdline.html#cmdoption-X
[PYTHON_CPU_COUNT]: https://docs.python.org/3.13/using/cmdline.html#envvar-PYTHON_CPU_COUNT
[CPU_COUNT]: https://docs.python.org/3.13/library/os.html#os.cpu_count

#### The issue

pytest-xdist does not explicitly respect these standard Python override mechanisms in its `pytest_xdist_auto_num_workers` hook. Assuming that `PYTEST_XDIST_AUTO_NUM_WORKERS` is unset:

- If `psutil` is available, then pytest-xdist will dismiss `-X cpu_count`/`PYTHON_CPU_COUNT`, because `psutil` uses C code to calculate the true logical process count available to this Python process.[^1]
- Otherwise, `os.sched_getaffinity` will be queried, which also dismisses `-X cpu_count`/`PYTHON_CPU_COUNT`.

#### The question

Should these override mechanisms be respected?

- I would argue yes, because they clearly state intent (beyond just the pytest-xdist interaction) for Python to <del>lie</del> <ins>claim</ins> to the script that only so-and-so many CPU cores are available.
- However, the code was written with pre-3.13 semantics in mind, and still runs on pre-3.13 systems, so retrofitting these overrides there is a compatibility break, and should perhaps not be taken lightly.
- Or perhaps the correct choice here is to go version-specific, i.e., to only respect the overrides on 3.13 and higher.

#### Going forward

If the override mechanisms are to be respected (whether in general, or on 3.13 and higher only), I would be willing to write a PR.

[^1]: It makes sense that `psutil` – which aims to query and report true system properties – would not respect these Python-specific overrides.

the-13th-letter:
Support `-X cpu_count` and `PYTHON_CPU_COUNT` in the `pytest_xdist_auto_num_workers` hook, on all Python versions. The `PYTEST_XDIST_AUTO_NUM_WORKERS` environment variable overrides `-X cpu_count`, which in turn overrides `PYTHON_CPU_COUNT`, which in turn inhibits querying `psutil`.

On Python 3.13, also support the `os.process_cpu_count` function in case `-n logical` is requested. This overrides querying `psutil`, and is in turn overridden by `PYTHON_CPU_COUNT`.

Fixes #1296

---

Open questions:

- As of Python 3.13, the documentation for `os.cpu_count` and `os.process_cpu_count` only speaks of logical CPUs. Should the xdist documentation (or even the implementation?) similarly retire the distinction between physical and logical CPUs?
- Should the documentation prominently mention the xdist version from which the behavior change (`-X` option/`PYTHON_CPU_COUNT`) is available?

Other feedback wanted:

- Is the "mood" of the documentation (and commentary, and changelog entry, etc.) correct? There was no concrete guidance, so I'm going by whatever I think is sensible…

---

Checklist:

- [x] We use [towncrier](https://pypi.org/project/towncrier) for changelog management, so please add a *news* file into the `changelog` folder following these guidelines:

    ```
    ```

### DIFF
--- a/changelog/1296.feature.rst
+++ b/changelog/1296.feature.rst
@@ -1,0 +1,5 @@
+``-n auto`` and ``-n logical`` now support the ``PYTHON_CPU_COUNT``
+environment variable and the ``-X cpu_count`` option from Python 3.13 (but
+on all supported Python versions) to override the auto-detected CPU count.
+The ``PYTEST_XDIST_AUTO_NUM_WORKERS`` environment variable has precedence
+over these.
--- a/docs/distribution.rst
+++ b/docs/distribution.rst
@@ -14,8 +14,10 @@ With ``-n auto``, pytest-xdist will use as many processes as your computer
 has physical CPU cores.
 
 Use ``-n logical`` to use the number of *logical* CPU cores rather than
-physical ones. This currently requires the `psutil <https://pypi.org/project/psutil/>`__ package to be installed;
-if it is not or if it fails to determine the number of logical CPUs, fall back to ``-n auto`` behavior.
+physical ones. This currently requires either Python 3.13 or higher, or the
+`psutil <https://pypi.org/project/psutil/>`__ package to be installed. If
+neither method is available or if they all fail to determine the number of
+logical CPUs, fall back to ``-n auto`` behavior.
 
 Pass a number, e.g. ``-n 8``, to specify the number of processes explicitly.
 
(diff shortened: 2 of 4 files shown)


RonnyPfannschmidt (approved):
Looks good

The logic in that one function is perhaps big enough to make a split sensible 

Apologies for the delays im stretched

septatrix:
As a result of this, should psutil maybe be dropped as an optional dependency based on the python version? (I.e. using the `python_version` marker for that dependency in `pyproject.toml`)

Issue #1248, opened 2025-09-23
`TestLoadScope::test_workqueue_ordered_by_input` failing intermittently


sliedes:
Maybe related: #985 

I'm doing mass rebuilds of packages in NixOS, with `-D_GLIBCXX_DEBUG` passed to compilations that link to libstdc++. The latter fact may or may not be relevant.

I am seeing intermittent failures of `TestLoadScope::test_workqueue_ordered_by_input` in pytest-xdist. It sometimes succeeds, sometimes not, probably depending on things like CPU load.

When it fails, it shows this failure:

```
________________ TestLoadScope.test_workqueue_ordered_by_input _________________
[gw23] linux -- Python 3.13.7 /nix/store/aj70vbwps8vslhww4p2glmi8jlbr3hp8-python3-3.13.7/bin/python3.13

self = <acceptance_test.TestLoadScope object at 0x7ffff54f0770>
pytester = <Pytester PosixPath('/build/pytest-of-nixbld/pytest-0/popen-gw23/test_workqueue_ordered_by_input0')>

    def test_workqueue_ordered_by_input(self, pytester: pytest.Pytester) -> None:
        test_file = """
            import pytest
            @pytest.mark.parametrize('i', range({}))
            def test(i):
                pass
        """
        pytester.makepyfile(test_a=test_file.format(10), test_b=test_file.format(20))
        result = pytester.runpytest(
            "-n2", "--dist=loadscope", "--no-loadscope-reorder", "-v"
        )
>       assert get_workers_and_test_count_by_prefix(
            "test_a.py::test", result.outlines
        ) == {"gw0": 10}
E       AssertionError: assert {'gw1': 10} == {'gw0': 10}
E
E         Left contains 1 more item:
E         {'gw1': 10}
E         Right contains 1 more item:
E         {'gw0': 10}
E         Use -v to get more diff

/build/pytest_xdist-3.8.0/testing/acceptance_test.py:1268: AssertionError
```

Issue #985, opened 2023-12-09
test_workqueue_ordered_by_size is flaky


mweinelt:
After upgrading pytest-xdist from 3.3.1 to 3.5.0 we are seeing the following test failure come up once in a while.

```
============================= test session starts ==============================
platform linux -- Python 3.11.6, pytest-7.4.3, pluggy-1.3.0
rootdir: /build/pytest-xdist-3.5.0
configfile: tox.ini
testpaths: testing
plugins: xdist-3.5.0
24 workers [195 items]   m
s.x......................x.x...........................s................ [ 36%]
.........x................s..................F......x........x.xx....... [ 73%]
.....x.......s...........s..s............x..x......                      [100%]
=================================== FAILURES ===================================
_________________ TestLoadScope.test_workqueue_ordered_by_size _________________
[gw14] linux -- Python 3.11.6 /nix/store/myhqvbw8sknili2cn9av3pncg63rkd0p-python3-3.11.6/bin/python3.11

self = <acceptance_test.TestLoadScope object at 0x7ffff58baf10>
pytester = <Pytester PosixPath('/build/pytest-of-nixbld/pytest-0/popen-gw14/test_workqueue_ordered_by_size0')>

    def test_workqueue_ordered_by_size(self, pytester: pytest.Pytester) -> None:
        test_file = """
            import pytest
            @pytest.mark.parametrize('i', range({}))
            def test(i):
                pass
        """
        pytester.makepyfile(test_a=test_file.format(10), test_b=test_file.format(20))
        result = pytester.runpytest("-n2", "--dist=loadscope", "-v")
>       assert get_workers_and_test_count_by_prefix(
            "test_a.py::test", result.outlines
        ) == {"gw1": 10}
E       AssertionError: assert {'gw0': 10} == {'gw1': 10}
E         Left contains 1 more item:
E         {'gw0': 10}
E         Right contains 1 more item:
E         {'gw1': 10}
E         Use -v to get more diff

/build/pytest-xdist-3.5.0/testing/acceptance_test.py:1244: AssertionError
```

PR #1318, opened 2026-03-31
fix: ensure consistent node workload assignment by sorting nodes by gwid

ttimonen:
Fixes the flaky test_workqueue_ordered_by_size test (#985 and #1248 )

- [x] We use [towncrier](https://pypi.python.org/pypi/towncrier) for changelog management, so please add a *news* file into the `changelog` folder.

### DIFF
--- a/changelog/985.bugfix.rst
+++ b/changelog/985.bugfix.rst
@@ -1,0 +1,1 @@
+De-flaked test_workqueue_ordered_by_size.
--- a/src/xdist/scheduler/loadscope.py
+++ b/src/xdist/scheduler/loadscope.py
@@ -394,11 +394,12 @@ class LoadScopeScheduling:
                 unused_node.shutdown()
 
         # Assign initial workload
-        for node in self.nodes:
+        # Sort the nodes to keep assingment consistent (#985)
+        for node in sorted(self.nodes, key=lambda node: node.gateway.id):
             self._assign_work_unit(node)
 
         # Ensure nodes start with at least two work units if possible (#277)
-        for node in self.nodes:
+        for node in sorted(self.nodes, key=lambda node: node.gateway.id):
             self._reschedule(node)
 
         # Initial distribution sent all tests, start node shutdown


RonnyPfannschmidt (commented):
That one is interesting 

Ordered at that particular place shouldn't be necessary 

We may have a timing issue in worker startup

ttimonen:
> That one is interesting
> 
> Ordered at that particular place shouldn't be necessary
> 
> We may have a timing issue in worker startup

Let me try to expand on what I think happens here, based on my understanding.

- call to test_workqueue_ordered_by_size(...) => 
1. We create test_a.py and test_b.py.
2. We invoke: pytest -n2 --dist=loadscope test_a.py test_b.py. A lot of things happen.

from xdist’s perspective:
a) pytest_runtestloop() is called, triggering creation of LoadScopeScheduling() object.
b) worker_workerready() is called, triggering a call to sched.add_node().
c) worker_collectionfinish() is called, triggering a call to sched.schedule().
Didn’t verify, but I’m assuming ordering of a,b,c. 

Now the interesting part is probably on the event processing in the workerready -side. So in b the add_node adds the nodes to the assigned_work dictionary in arbitrary order. While python dicts are all ordered nowadays, the insertion order varies here. I.e. add_node calls come from the DSession’s event queue, and the two gw remotes are competing in which one will submit the “workerready” event first (irrespective of the order they were started). So add_node(gw0) and add_node(gw1) will be called in arbitrary order.
This is ok in isolation. However…

3. output of 2. has lines like
3a:
[gw0] [  3%] PASSED test_b.py::test[0] 
[gw1] [  6%] PASSED test_a.py::test[0] 
...
or 3b
[gw1] [  3%] PASSED test_b.py::test[0] 
[gw0] [  6%] PASSED test_a.py::test[0] 
..
depending on which worker became ready first.

To make output deterministic, we could either synthesize the gw identity based on their order of becoming ready (might have other side-effects though), assign work on name-order (which is done with this patch and seems like the least amount of friction), or do something else (like hide the id from output, since it’s semantically irrelevant (yet part of the implicit API now), or make the test ignore the gw id; bit ambiguous what the test would be testing in that case though).

ttimonen:
A minimal example to reproduce the problem and the effectiveness of the fix:

```
[ testing]$ nix shell --impure --expr 'let p = import <nixpkgs> {}; in [ p.nushell (p.python3.withPackages (x: with x; [ pytest execnet pytest-xdist ])) ]' \
  -c nu  -c '1..100 | par-each {pytest acceptance_test.py -k test_workqueue_ordered_by_size | complete} | group-by --to-table exit_code | update items {|x| $x.items | length}'
╭───┬───────────┬───────╮
│ # │ exit_code │ items │
├───┼───────────┼───────┤
│ 0 │ 1         │    31 │
│ 1 │ 0         │    69 │
╰───┴───────────┴───────╯
```
Now replace the default pytest-xdist from current branch with PYTHONPATH=../src
```
[ testing]$ nix shell --impure --expr 'let p = import <nixpkgs> {}; in [ p.nushell (p.python3.withPackages (x: with x; [ pytest execnet pytest-xdist ])) ]' \
  -c nu  -c '1..100 | par-each {PYTHONPATH=../src pytest acceptance_test.py -k test_workqueue_ordered_by_size | complete} | group-by --to-table exit_code | update items {|x| $x.items | length}'
╭───┬───────────┬───────╮
│ # │ exit_code │ items │
├───┼───────────┼───────┤
│ 0 │ 0         │   100 │
╰───┴───────────┴───────╯
```

RonnyPfannschmidt (approved):

RonnyPfannschmidt:
Thanks for the followup im aiming for merge/ release next week

PR #1324, opened 2026-04-18
Fix hang with `--dist=loadgroup` if a crashed worker is replaced

### LINKED ISSUE #1323: `pytest --dist=loadgroup` hangs if a crashed worker is restarted
When running the following test file

```python
def test_1():
    assert True

def test_2():
    import time
    time.sleep(5)
    assert True
```

with (uses features from pytest 9)

```bash
pytest -n1 --dist=loadgroup -o faulthandler_timeout=1 -o faulthandler_exit_on_timeout=true testing/test_timeout.py
```

test execution hangs with

```
replacing crashed worker gw0
collecting: 1/2 workers
collecting: 1/2 workers
2 workers [2 items]
```

When using another test distribution algorithm, test execution does not hang.

The issue is that https://github.com/pytest-dev/pytest-xdist/blob/8fed345a98ac673dd41e52b2b5f2a9783c58d3dc/src/xdist/scheduler/loadscope.py#L199-L200
adds the complete workload to the queue, including the completed work.

Later in https://github.com/pytest-dev/pytest-xdist/blob/8fed345a98ac673dd41e52b2b5f2a9783c58d3dc/src/xdist/scheduler/loadscope.py#L276-L280
completed work items are dropped so that `nodeids_indexes` is empty.

Finally, the process hangs in https://github.com/pytest-dev/pytest-xdist/blob/8fed345a98ac673dd41e52b2b5f2a9783c58d3dc/src/xdist/remote.py#L204

radoering:
Resolves #1323

In addition to the fix in `remove_node()`, I added a sanity check in `_assign_work_unit()`. (Without the fix and the sanity check the unit test hangs. Without the fix but with the sanity check, it fails due to the `RuntimeError`.)

### DIFF
--- a/changelog/1323.bugfix.rst
+++ b/changelog/1323.bugfix.rst
@@ -1,0 +1,1 @@
+Fix hang with `--dist=loadgroup` if a crashed worker is replaced.
--- a/src/xdist/scheduler/loadscope.py
+++ b/src/xdist/scheduler/loadscope.py
@@ -196,8 +196,10 @@ class LoadScopeScheduling:
                 "Unable to identify crashitem on a workload with pending items"
             )
 
-        # Made uncompleted work unit available again
-        self.workqueue.update(workload)
+        # Make uncompleted work units available again
+        for scope, work_unit in workload.items():
+            if any(not completed for completed in work_unit.values()):
+                self.workqueue[scope] = work_unit
 
         for node in self.assigned_work:
             self._reschedule(node)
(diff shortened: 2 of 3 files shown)


RonnyPfannschmidt (approved):
Thanks

radoering:
@RonnyPfannschmidt Since you approved the PR some time ago, I am wondering what is still preventing the merge?

radoering:
@RonnyPfannschmidt Thanks for merging. :pray: Maybe, if time permits you can take a look if #1328 is also an acceptable fix? (Different root cause, but similar consequences.)

PR #1328, opened 2026-04-27
Fix hang with `--dist=loadgroup` if a crashed worker is replaced and there is still work to be done

### LINKED ISSUE #1327: `pytest --dist=loadgroup` hangs if a crashed worker is restarted - variant 2
This issue is similar to #1323 (regarding the issue that can be observed from the outside), but since it is triggered by a slightly different setup, the root cause is different and it requires a different fix, I created a new issue.

When running the following test file

```python
def test_1():
    import time
    time.sleep(5)
    assert True

def test_2():
    assert True
```

with

```bash
pytest -n1 --dist=loadgroup -o faulthandler_timeout=1 -o faulthandler_exit_on_timeout=true testing/test_timeout.py
```

test execution hangs with

```
replacing crashed worker gw0
2 workers [2 items]
```

The issue is that only one item is assigned to the new worker in https://github.com/pytest-dev/pytest-xdist/blob/8fed345a98ac673dd41e52b2b5f2a9783c58d3dc/src/xdist/scheduler/loadscope.py#L352-L355

This item is removed from the queue in https://github.com/pytest-dev/pytest-xdist/blob/8fed345a98ac673dd41e52b2b5f2a9783c58d3dc/src/xdist/remote.py#L204

Then, in https://github.com/pytest-dev/pytest-xdist/blob/8fed345a98ac673dd41e52b2b5f2a9783c58d3dc/src/xdist/remote.py#L214

the worker waits for a second item.

In contrast, at first start (when `self.collection` is still `None`), two items are added to the queue via https://github.com/pytest-dev/pytest-xdist/blob/8fed345a98ac673dd41e52b2b5f2a9783c58d3dc/src/xdist/scheduler/loadscope.py#L396-L402

radoering:
Resolves #1327

### DIFF
--- a/changelog/1327.bugfix.rst
+++ b/changelog/1327.bugfix.rst
@@ -1,0 +1,1 @@
+Fix hang with `--dist=loadgroup` if a crashed worker is replaced.
--- a/src/xdist/scheduler/loadscope.py
+++ b/src/xdist/scheduler/loadscope.py
@@ -352,6 +352,12 @@ class LoadScopeScheduling:
         if self.collection is not None:
             for node in self.nodes:
                 self._reschedule(node)
+            # Ensure nodes have at least two work units if possible,
+            # since workers need a "next item" before running the current one.
+            # (A restarted worker has no item before calling _reschedule()
+            # for the first time.)
+            for node in self.nodes:
+                self._reschedule(node)
             return
 
         # Check that all nodes collected the same tests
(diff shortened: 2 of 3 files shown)


RonnyPfannschmidt (approved):
this workaround points to an need to rework sheduling a little better

larsoner:
I can confirm that this fixes the crashes from #1327

PR #1330, opened 2026-05-10
feat: add worker ramp option

Harshal96:
## Summary

Add `--ramp DURATION` to stagger when xdist workers begin executing tests after collection.

Refs #1219.

## Checklist

- [x] Added towncrier changelog entry: `changelog/1219.feature.rst`

## Test Plan

- `uv run --extra testing pytest testing/test_plugin.py testing/test_workermanage.py testing/test_remote.py testing/acceptance_test.py -q`
- `uv run --python 3.10 --with pre-commit pre-commit run --files ...`

### DIFF
--- a/src/xdist/plugin.py
+++ b/src/xdist/plugin.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import math
 import os
 import sys
 from typing import TYPE_CHECKING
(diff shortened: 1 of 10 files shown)


RonnyPfannschmidt:
This deviates from then proposed mechanism as it sidesteps the start later detail 

As such it's a partial solution to the proposal but also a working solution that has to solve far less problems than a full process ramping solution 

Its a pretty nice step forward

Harshal96:
> This deviates from then proposed mechanism as it sidesteps the start later detail
> 
> As such it's a partial solution to the proposal but also a working solution that has to solve far less problems than a full process ramping solution
> 
> Its a pretty nice step forward

@RonnyPfannschmidt any suggested changes to make it an acceptable contribution?

RonnyPfannschmidt:
It already is

Harshal96:
@RonnyPfannschmidt awesome, this is my first contribution in this repo, what's the approval/merge process? Anything pending on me?

src/xdist/plugin.py
```diff
@@ -92,15 +93,13 @@ def _auto_num_workers_psutil(config: pytest.Config) -> int | None:
...
-
+    sched_getaffinity = getattr(os, "sched_getaffinity", None)
+    if sched_getaffinity is not None:
         return len(sched_getaffinity(0))
-    except ImportError:
-        if os.environ.get("TRAVIS") == "true":
-            # workaround https://github.com/pypy/pypy/issues/2375
-            return 2
-        return None
+    if os.environ.get("TRAVIS") == "true":
```
RonnyPfannschmidt, line 99: followup note - travis is no longer a thing in direct use - we might want to drop explicit handling for it

RonnyPfannschmidt (approved):
the workermanage api change is okay as its a internal class

RonnyPfannschmidt:
sorry for the delays, i'm a bit stretched

Harshal96:
@RonnyPfannschmidt quick reminder for merging this

Pierre-Sassoulas (approved):

PR #1340, opened 2026-06-14
Fix CLI options ignored when --looponfail runs with -n (#767)

### LINKED ISSUE #767: `--tb=short` option ignored when combined with `-n2` and `--looponfail`
Pytest's `--tb=short` option is ignored when combined with the `--looponfail` and `-n2` options of `pytest-xdist`.

#### To reproduce:
```python
# foo.py
def a():
    b()

def b():
    c()

def c():
    assert False
```
```python
# test_foo.py
from foo import a
def test_foo():
    a()
```
```commandline
$ pytest --tb=short -n2 --looponfail test_foo.py
...  # long-style traceback displayed, --tb=short option ignored.
```

golikovichev:
Fixes #767.

## Problem

`pytest --tb=short -n2 --looponfail` ignores `--tb=short` and prints a long traceback. The option is only dropped when `--looponfail` and `-n` are combined; each on its own respects `--tb=short`.

## Root cause

`--looponfail` rebuilds the worker session config via `Config.fromdictargs(option_dict, args)`, where `args` is the positional test paths only. `fromdictargs` sets `config.invocation_params.args` to exactly those positional arguments, so the original command-line flags (e.g. `--tb=short`) are not present there, even though `config.option` carries them.

That is fine for looponfail's own in-process run, which reads `config.option` directly. But when looponfail runs a nested distributed (`-n`) session, `WorkerController.setup` builds the nested worker argv from `config.invocation_params.args` (workermanage.py). Since the flags are missing there, the nested workers fall back to their defaults, including the long traceback style.

## Fix

Pass the original invocation arguments to the looponfail worker and restore a faithful `config.invocation_params.args` after `fromdictargs`. This propagates all command-line flags (not just `--tb`) to nested workers. `dataclasses.replace` is used because `InvocationParams` is a frozen dataclass; `plugins` and `dir` are preserved.

## Tests

Added a regression test in `testing/test_looponfail.py` that runs the failing tests through `RemoteControl.loop_once()` with `-n2 --tb=short` and asserts the short-style traceback markers appear (and the long-style source expansion does not).

Verified locally:
- New test fails without the fix, passes with it.
- Full `testing/test_looponfail.py` passes (no regressions).
- `ruff check`, `ruff format --check`, and `mypy` are clean.
- The original reproduction command from the issue now prints a short traceback.

### DIFF
--- a/changelog/767.bugfix.rst
+++ b/changelog/767.bugfix.rst
@@ -1,0 +1,5 @@
+Command-line options such as ``--tb=short`` are no longer dropped when
+``--looponfail`` is combined with distributed runs (``-n``). Previously the
+worker session was rebuilt only from the parsed options and positional
+arguments, so the original invocation flags were lost and the nested workers
+fell back to their defaults (for example, the long traceback style).
--- a/src/xdist/looponfail.py
+++ b/src/xdist/looponfail.py
@@ -164,7 +164,6 @@ def init_worker_session(
     option_dict: dict[str, "Any"],  # noqa: UP037
     invocation_args: list[str],
 ) -> None:
-    import dataclasses
     import os
     import sys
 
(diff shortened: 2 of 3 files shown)


RonnyPfannschmidt:
this needs a deep reviews - ithis trops on the incorrect config restore, but i take note that this only fixes part of the issue - im at the pytest sprint and will experiment with a more nuanced approach

golikovichev:
Thanks for looking, and no rush - enjoy the sprint.

You're right that it's partial. It targets the specific #767 case where `-n` + `--looponfail` drops the CLI flags on the nested workers: `fromdictargs` sets `invocation_params.args` to just the positional paths, so `workermanage` rebuilds the worker argv without the original flags. My change restores `invocation_params.args` after `fromdictargs` so they propagate.

If your read is that the restore is papering over an incorrect `fromdictargs` behaviour rather than fixing it at the source, I'm on board with a more nuanced approach - that path did feel like a symptom fix. Happy to pare this down to just the regression test in `test_looponfail.py` as a reproduction and let your version land the real fix, or rework it once you've settled a direction after the sprint. Whatever avoids churn against what you're experimenting with - just point me at it.

PR #1355, opened 2026-07-15
docs: document terminal-size inconsistency under -n

### LINKED ISSUE #1208: stdout width inconsistent
When running the tests of [Fandango](https://github.com/fandango-fuzzer/fandango), specifically `test_cli.test_help`, I'm getting a weird error. The test looks as follows:

```python
import os
import shlex
import shutil
import subprocess
import unittest

from fandango.cli import get_parser


class test_cli(unittest.TestCase):

    def run_command(self, command):
        proc = subprocess.Popen(
            command,
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE,
        )
        out, err = proc.communicate()
        return out.decode(), err.decode(), proc.returncode

    def test_help(self):
        command = shlex.split("fandango --help")
        out, err, code = self.run_command(command)
        parser = get_parser(True)
        self.maxDiff = 100000
        self.assertEqual(0, code)
        self.assertEqual(out, parser.format_help())
        self.assertEqual(err, "")
```

Now, if I run `pytest -k test_help`, the test succeeds. If I however specify `-n {1,auto}`, I get the following error:

```
E       AssertionError: 'usag[45 chars]uiet] {fuzz,parse,shell,help,copyright,version[927 chars]n.\n' != 'usag[45 chars]uiet]\n                {fuzz,parse,shell,help,[944 chars]n.\n'
E       - usage: fandango [-h] [--version] [--verbose | --quiet] {fuzz,parse,shell,help,copyright,version} ...
E       + usage: fandango [-h] [--version] [--verbose | --quiet]
E       +                 {fuzz,parse,shell,help,copyright,version} ...
E         
E         The access point to the Fandango framework
E         
E         options:
E           -h, --help            show this help message and exit
E           --version             show version number
E           --verbose, -v         increase verbosity. Can be given multiple times (-vv)
E           --quiet, -q           decrease verbosity. Can be given multiple times (-qq)
E         
E         commands:
E           {fuzz,parse,shell,help,copyright,version}
E                                 the command to execute
E             fuzz                produce outputs from .fan files and test programs
E             parse               parse input file(s) according to .fan spec
E             shell               run an interactive shell (default)
E             help                show this help and exit
E             copyright           show copyright
E             version             show version
E         
E         Use `fandango help` to get a list of commands.
E         Use `fandango help COMMAND` to learn more about COMMAND.
E         See https://fandango-fuzzer.github.io/ for more information.
```

Seems like the stdout width is different between the two implementations? I assume this is a bug?

pctablet505:
Fixes #1208

Documents that pytest-xdist workers run with replaced stdio pipes, so code that queries the terminal width sees the default size, and suggests monkeypatching the expected width in tests.

### DIFF
--- a/changelog/1208.doc.rst
+++ b/changelog/1208.doc.rst
@@ -1,0 +1,1 @@
+Document that workers run with replaced standard streams and a default terminal size, so tests asserting on width-dependent output may need to monkeypatch the expected width.
--- a/docs/known-limitations.rst
+++ b/docs/known-limitations.rst
@@ -69,3 +69,13 @@ Debugging
 This also means that debugging using PDB (or any other debugger that wants to use standard I/O) will not work. The ``--pdb`` option is disabled when distributing tests with ``pytest-xdist`` for this reason.
 
 It is generally likely best to use ``pytest-xdist`` to find failing tests and then debug them without distribution; however, if you need to debug from within a worker process (for example, to address failures that only happen when running tests concurrently), remote debuggers (for example, `python-remote-pdb <https://github.com/ionelmc/python-remote-pdb>`__ or `python-web-pdb <https://github.com/romanvm/python-web-pdb>`__) have been reported to work for this purpose.
+
+Terminal size
+-------------
+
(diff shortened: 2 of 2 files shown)


RonnyPfannschmidt (approved):

RonnyPfannschmidt:
thanks

PR #1348, opened 2026-06-27
fix: worker restart collection mismatch in loadscope/loadgroup scheduler

C1-BA-B1-F3:
## Description

This PR fixes two related issues when a worker crashes and is restarted with `--dist loadscope` or `--dist loadgroup`:

### Issue #1189: KeyError when replacement worker's collection doesn't match

When a worker crashes, its entry in `registered_collections` was not cleaned up. This caused:
1. `collection_is_completed` to remain True even though the crashed node was gone
2. Replacement workers whose collections didn't match the original would have their collections silently dropped (the `return` statement in `add_node_collection()`)
3. When `schedule()` tried to assign work to the replacement worker, it would raise `KeyError` because the worker wasn't in `registered_collections`

### Issue #1323: Hang when completed work units are requeued

When a worker crashed, ALL work units (including completed ones) were added back to the workqueue via `self.workqueue.update(workload)`. When these completed work units were later assigned to a new worker:
1. The filtering in `_assign_work_unit()` would produce an empty `nodeids_indexes` list (all items marked as completed)
2. `node.send_runtest_some([])` would be called with an empty list
3. The worker would hang waiting for work that never arrives

### `src/xdist/scheduler/loadscope.py`

1. **`remove_node()`**: Now removes the crashed node from `registered_collections` using `self.registered_collections.pop(node, None)`. This allows replacement workers to properly register their collections.

2. **`remove_node()`**: Only requeues work units that have pending (uncompleted) items. Completed work units are filtered out before adding to the workqueue.

3. **`add_node_collection()`**: Always registers the collection even if it doesn't match the existing collection. The mismatch will be properly reported by `_check_nodes_have_same_collection()` when `schedule()` is called.

## Testing

All existing tests pass (220 passed, 6 skipped, 10 xfailed).

## Related Issues

Fixes #1189
Fixes #1323

PR #1363, opened 2026-07-25
Fix loadscope scheduler deadlock after a worker crash

larsoner:
- [x] We use [towncrier](https://pypi.python.org/pypi/towncrier) for changelog management, so please add a *news* file into the `changelog` folder ...

I used Claude Fable 5 to look into https://github.com/pytest-dev/pytest-xdist/issues/1313 (after hitting a very reliable hang in MNE-Python in https://github.com/mne-tools/mne-python/pull/14063 during a ~20m run). With some guidance, Claude was able to boil it down into a minimal repro (now adapted into a unit test) and suggest this fix based on its conclusion:

> The bug is that `mark_test_complete()` never prunes fully-completed work units from `assigned_work`, so when a worker crashes, remove_node() requeues everything it ever ran. Replacement (or endgame) workers then get assigned an already-completed unit as an empty runtests command, which can never be marked complete — so _reschedule() never fires again and the session deadlocks with idle workers.

I then tested in https://github.com/mne-tools/mne-python/pull/14101 that 1) it still hung in MNE-Python with latest pytest-xdist `master` (it [did](https://github.com/mne-tools/mne-python/actions/runs/30124895924/job/89585961010)), then 2) used this PR and saw [it succeed in ending properly after an error rather than hanging](https://github.com/mne-tools/mne-python/actions/runs/30126505675/job/89591159426). It coming back red allowed me to narrow down the bug to a Qt6<->Cocoa issue (with multiple runs, each ending properly on this branch) and work around it to [get to green](https://github.com/mne-tools/mne-python/actions/runs/30134978477/job/89616908392) 🎉 !

Related to https://github.com/pytest-dev/pytest-xdist/issues/1313 and addressed what I commented on there, but not sure if it actually closes it or not. @clemlesne you could try my branch and check if you have the spare cycles (I pushed upstream tags to git so the dynamic dev version gets resolved correctly, and used `pip install "pytest-xdist @ git+https://github.com/larsoner/pytest-xdist.git@lock"` in my own testing.)


larsoner:
A minor status update here / some evidence that it's working -- we've been using this branch for the last month in MNE-Python without issue (and it works correctly when we've had an occasional segfault that previously would have hung).

larsoner:
Actually I think #1328 is a better fix, I'll close!

larsoner:
(and #1348 looks still useful, too)

PR #1299, opened 2026-01-23
Fix `KeyError` when replacement worker has mismatched collection

bendrucker:
Fixes a `KeyError` crash when a replacement worker's test collection doesn't match the original. This can occur when using `--dist=loadgroup` and a worker crashes, then the replacement worker sees different tests (e.g., due to a race condition or test file changes during the run).

## Changes

- Add a guard in `_reschedule()` to skip nodes that aren't in `registered_collections`
- Add regression test that simulates the crash scenario

## Root Cause

When a worker crashes and is replaced:
1. The replacement worker collects tests via `add_node_collection()`
2. If the collection doesn't match the original, the method [returns early](https://github.com/pytest-dev/pytest-xdist/blob/04c3ca8b9f6ade5e8ccabca9fa5b7d9e65140c0d/src/xdist/scheduler/loadscope.py#L234) **without** adding the node to `registered_collections`
3. However, the node **was** already added to `assigned_work` via `add_node()`
4. Later, `schedule()` iterates over `self.nodes` (from `assigned_work.keys()`) and calls `_reschedule()` on each
5. `_reschedule()` calls `_assign_work_unit()` which [crashes with `KeyError`](https://github.com/pytest-dev/pytest-xdist/blob/04c3ca8b9f6ade5e8ccabca9fa5b7d9e65140c0d/src/xdist/scheduler/loadscope.py#L275)

The fix adds a [simple guard](https://github.com/pytest-dev/pytest-xdist/blob/78bec67f82898e04a6492717ffd109cd2fa7e96e/src/xdist/scheduler/loadscope.py#L323-L326) to skip nodes that aren't properly registered.

## Related

- Originally observed in: https://github.com/pydantic/pydantic-ai/actions/runs/21232509180/job/61093663173?pr=3772
- Fixes #1189
- Fixes #714

PR #1364, opened 2026-07-26
loadscope: forget a dead worker's collection so a still-collecting worker isn't mistaken for done

### LINKED ISSUE #1362: LoadScopeScheduling.remove_node leaves a dead worker in registered_collections, causing INTERNALERROR KeyError when another worker is still collecting
### Summary

`LoadScopeScheduling.remove_node()` removes a dead worker from `assigned_work` but never removes its entry from `registered_collections`. The stale entry keeps `collection_is_completed` returning `True` while a *later* worker is still collecting, so `schedule()` proceeds and `_assign_work_unit()` raises `KeyError` on the still-collecting node — crashing the whole run with `INTERNALERROR`.

This affects both `--dist=loadscope` and `--dist=loadfile`, since `LoadFileScheduling` subclasses `LoadScopeScheduling`.

### Mechanism

In `xdist/scheduler/loadscope.py` (3.8.0):

- `collection_is_completed` → `len(self.registered_collections) >= self.numnodes`
- `remove_node()` → `workload = self.assigned_work.pop(node)` … and no corresponding removal from `self.registered_collections`

So after a worker dies, `registered_collections` still holds its entry. If a *different* worker has not yet finished collecting, `collection_is_completed` is already satisfied by the dead worker's stale entry. `worker_collectionfinish` then calls `schedule()`, whose `assert self.collection_is_completed` passes, and `_reschedule` → `_assign_work_unit` does `self.registered_collections[node]` for a node that has not registered → `KeyError`.

Traceback path: `dsession.py` `worker_collectionfinish` → `loadscope.py` `schedule` → `_reschedule` → `_assign_work_unit`.

### Observed

Two independent runs on the same repository, `--dist=loadfile -n auto` (24 workers on a 24-core host under heavy concurrent load):

```
[gw7] node down: Not properly terminated
[gw6] node down: Not properly terminated
[gw8] node down: Not properly terminated
INTERNALERROR> KeyError: <WorkerController gw14>
```

```
[gw2] node down: Not properly terminated
INTERNALERROR> KeyError: <WorkerController gw12>
```

Note the pattern in both: the workers that **die** are low-index, and the `KeyError` names a **high-index** worker — i.e. one that had not yet finished collecting. That is exactly what the stale-entry mechanism predicts.

### Secondary symptom: duplicated reports for a single node

In a separate run on the same codebase, a module collecting 6 tests produced 11 reports, with one non-parametrised test listed **6 times** in the FAILED block, after 5 workers went down and their work was rescheduled onto successive workers. So beyond the crash, bookkeeping desynchronisation appears able to inflate report counts — a FAILED block from a run that lost workers cannot be read as a failure inventory.

### Impact

The run exits 3, so CI does fail correctly — this is not a false green. The damage is diagnostic: no FAILED summary is printed, counts are partial or inflated, and the remaining tests never execute. The only recourse is a full re-run (20–45 minutes on our suite), which is how "just re-run it" becomes institutionalised and why failure sets are not comparable between runs.

Worker deaths themselves are environmental here (heavy host contention), and we are capping worker count as mitigation — but a scheduler crash on worker loss seems worth fixing independently, since `remove_node`'s own docstring states it is called "either when the node crashed or at shutdown time".

### Suggested fix

Remove the node's `registered_collections` entry in `remove_node()` alongside the `assigned_work.pop(node)`, so `collection_is_completed` reflects only live nodes. (I have not opened a PR as I have not evaluated the effect on `numnodes` accounting during shutdown, where `remove_node` is also called.)

### Environment

- pytest-xdist 3.8.0 (latest on PyPI at time of filing)
- pytest 9.1.1
- Python 3.13.11, Windows

dchaudhari7177:
Fixes #1362.

### Problem

`LoadScopeScheduling.remove_node()` pops the dead worker from `assigned_work` but never removes its entry from `registered_collections`. Since `collection_is_completed` is `len(registered_collections) >= numnodes`, the stale entry keeps the collection looking "complete" while a *later* worker is still collecting. `schedule()` then runs and `_assign_work_unit()` raises `KeyError` indexing `registered_collections` for the worker that never registered — crashing the run with `INTERNALERROR`. This affects both `--dist=loadscope` and `--dist=loadfile` (the latter subclasses `LoadScopeScheduling`).

### Fix

Drop the node from `registered_collections` in `remove_node()`, mirroring the `assigned_work.pop(node)` already there. A replacement worker then re-registers normally, and `collection_is_completed` correctly waits for every live worker.

This is safe for the shutdown path too: normal `worker_workerfinished` removals only happen after `triggershutdown`, and on a crash the replacement worker re-registers its collection.

### Test

Added `TestLoadScopeScheduling::test_remove_node_forgets_dead_worker_collection`, which reproduces the scenario with `MockNode`s (3 expected workers; one dies mid-collection while another is still collecting). It **fails on `main`** (`registered_collections` still holds the dead node, `collection_is_completed` flips true early) and passes with the fix. Full `testing/test_dsession.py` is green; `ruff check`/`format` clean. Added `changelog/1362.bugfix.rst`.


larsoner:
Is this related to / dup of https://github.com/pytest-dev/pytest-xdist/pull/1363 ?

dchaudhari7177:
No — different bug, different line, and they can both be wrong at once. Same file, which is why they look alike.

**#1363 / #1313** is about `mark_test_complete` never pruning a *finished* work unit from `assigned_work`, so `remove_node` requeues it and a replacement worker gets an empty unit it can never complete. You closed it in favour of #1328, which fixes that class of hang for `loadgroup`.

**This PR / #1362** is about `remove_node` popping the dead node from `assigned_work` but leaving it in `registered_collections`. Since

```python
def collection_is_completed(self) -> bool:
    return len(self.registered_collections) >= self.numnodes
```

counts entries rather than live workers, the stale entry makes the collection look complete while a *later* worker is still collecting. `schedule()` then runs and `_assign_work_unit` raises `KeyError` indexing `registered_collections` for the worker that never registered. The failure mode is an `INTERNALERROR` traceback, not a hang.

The two diffs do not touch the same function — #1363 edits `mark_test_complete` (line ~252), this edits `remove_node` (line ~182) — and #1328, which landed instead, does not touch `registered_collections` at all. I re-read `remove_node` on `master` at `e27d3bb` to confirm: the `registered_collections.pop` is still absent, and `testing/test_dsession.py::TestLoadScopeScheduling::test_remove_node_forgets_dead_worker_collection` still fails there.

Happy to rebase and re-run if that helps.

dchaudhari7177:
Not a duplicate — different bug, different line, and they don't overlap.

#1363 (and #1328, which you preferred) is about #1313: a crashed worker's *completed* work units get requeued, a replacement is handed an empty unit, and the session hangs. It changes `mark_test_complete` and `_assign_work_unit`.

This one is #1362: `remove_node` pops the dead worker from `assigned_work` but leaves its entry in `registered_collections`. `collection_is_completed` is `len(registered_collections) >= numnodes`, so the stale entry keeps collection looking complete while a *later* worker is still collecting; `schedule()` then runs and `_assign_work_unit` raises `KeyError` on the worker that never registered, so the run dies with `INTERNALERROR` rather than hanging. The fix is one line in `remove_node`, mirroring the `assigned_work.pop(node)` next to it.

So a session could hit either one independently: #1328 doesn't touch `registered_collections`, and this doesn't touch requeueing. Happy to rebase if that helps.

davidheff:
Thanks for this. The stale entry is a real fault, and dropping it fixes the `schedule()` route. But I think a second route to the same `KeyError` remains with this change applied.

`remove_node` ends by calling `_reschedule` on every node in `assigned_work`. A replacement worker is in `assigned_work` as soon as `add_node` runs, so it's there before it has reported its collection. If a second worker crashes while the first replacement is still collecting, the second `remove_node` re-queues the dead worker's units and then reschedules the replacement. `_reschedule` sees nothing pending and calls `_assign_work_unit`, which raises `KeyError` looking up `registered_collections` for the replacement.

`worker_errordown` swallows `KeyError` from `remove_node` (`except KeyError: pass`), so this route doesn't take the run down the way the `schedule()` route does. But it skips `handle_crashitem`, so the test the second worker crashed on is never reported, and the loop stops before rescheduling any node after the replacement.

Reproduced against this branch (e6ef607) and against master with a real `LoadScopeScheduling` and mock nodes:

```python
import types
from xdist.scheduler.loadscope import LoadScopeScheduling

class Node:
    def __init__(self, name):
        self.gateway = types.SimpleNamespace(id=name)
        self.shutting_down = False
        self.sent = []
    def send_runtest_some(self, indices):
        self.sent.extend(indices)
    def shutdown(self):
        self.shutting_down = True

config = types.SimpleNamespace(
    getvalue=lambda name: ["3*popen"],
    option=types.SimpleNamespace(loadscopereorder=False),
)
collection = [f"test_{m}.py::test_{i}" for m in "abcdef" for i in range(3)]
sched = LoadScopeScheduling(config)
a, b, c = Node("gw0"), Node("gw1"), Node("gw2")
for node in (a, b, c):
    sched.add_node(node)
for node in (a, b, c):
    sched.add_node_collection(node, collection)
    if sched.collection_is_completed:
        sched.schedule()

sched.remove_node(a)   # gw0 crashes
sched.add_node(Node("gw3"))   # its replacement starts collecting
sched.remove_node(b)   # gw1 crashes: KeyError, on this branch and on master
```

What closes both routes for us is the check from #1299: `_reschedule` returns early for a node that isn't in `registered_collections` yet. It's then scheduled normally when its own collection arrives. With that in place, dropping the stale entry is still correct, since it keeps `collection_is_completed` honest, but it's no longer what prevents the `KeyError`. Would you consider adding the guard here, or should the two PRs be combined?

For context, we hit this in a real run when two workers died in the same second. The `INTERNALERROR` came from the `schedule()` route and terminated the entire run before it completed.

PR #1372, opened 2026-08-16
Transfer the warning's state, and rebuild without re-running `__init__`

### LINKED ISSUE #404: 'module' object has no attribute 'GtkWarning'
Hello. First of all I would like to thank everyone involved in making and maintaining pytest and xdist, superb job. :) 👍 

Now the situation that is manifesting on my windows7 machine.
I'm trying to assemble CI infrastructure using jenkins for running gui tests via pytest and xdist.
I decided to use xdist because python process sometimes crashes and I want to still continue running the remaining number of tests. This is where `-n` comes into play. I've managed to reproduce situation bellow also in the console (`cmd.exe`).

Bellow we can see one of the commands I'm using to run pytest, I've moved some args onto a new line so it is easier on the eyes.

```
C:\Users\user\.jenkins\workspace\Rmaster_GUI>(D: &&
D:/python_folder/python -m pytest --disable-pytest-warnings --showlocals -n1
--cov=D:/installation_folder/python/client/project_name
--cov-report xml:C:\Users\user\.jenkins\workspace\Rmaster_GUI/coverage-reports/coverage.xml
--junit-xml=C:\Users\user\.jenkins\workspace\Rmaster_GUI/test-reports/results.xml
-v
C:\Users\user\.jenkins\workspace\Rmaster_GUI/test/project_name_test )

17:54:47 ============================= test session starts =============================
17:54:47 platform win32 -- Python 2.7.10, pytest-4.1.1, py-1.5.4, pluggy-0.8.1 -- D:\python_folder\python.exe
17:54:47 cachedir: .pytest_cache
17:54:47 rootdir: D:\, inifile:
17:54:47 plugins: xdist-1.26.0, forked-1.0.1, cov-2.5.1
17:54:47 gw0 I
17:54:47 
[gw0] win32 Python 2.7.10 cwd: D:\
17:54:48 
[gw0] Python 2.7.10 (default, Nov 29 2018, 13:23:51) [MSC v.1900 64 bit (AMD64)]
17:55:45 gw0 [1096]
17:55:45 
17:55:45 scheduling tests via LoadScheduling

## After some time and a couple tests ...

17:59:22 [gw0] [  1%] PASSED integration/gui/test_some_component.py::TestSomeComponentEditor::test_component !!!!!!!!!!!!!!!!!!!! D:\python_folder\Lib\site-packages\xdist\workermanage.py:436: AttributeError: 'module' object has no attribute 'GtkWarning'
17:59:22 
17:59:22 INTERNALERROR> Traceback (most recent call last):
17:59:22 INTERNALERROR>   File "D:\python_folder\Lib\site-packages\xdist\workermanage.py", line 336, in process_from_remote
17:59:22 INTERNALERROR>     kwargs["warning_message_data"]
17:59:22 INTERNALERROR>   File "D:\python_folder\Lib\site-packages\xdist\workermanage.py", line 436, in unserialize_warning_message
17:59:22 INTERNALERROR>     cls = getattr(mod, data["message_class_name"])
17:59:22 INTERNALERROR> AttributeError: 'module' object has no attribute 'GtkWarning'
17:59:22 [gw0] node down: D:\python_folder\Lib\site-packages\xdist\workermanage.py:436: AttributeError: 'module' object has no attribute 'GtkWarning'
17:59:22 [gw0] [  1%] FAILED integration/gui/test_some_component.py::TestSomeComponentEditor::test_component
17:59:22 Replacing crashed worker gw0
17:59:22 
[gw1] win32 Python 2.7.10 cwd: D:\
17:59:22 Coverage.py warning: No data was collected. (no-data-collected)
17:59:37 INTERNALERROR> Traceback (most recent call last):
17:59:37 INTERNALERROR>   File "D:\python_folder\Lib\site-packages\_pytest\main.py", line 203, in wrap_session
17:59:37 INTERNALERROR>     session.exitstatus = doit(config, session) or 0
17:59:37 INTERNALERROR>   File "D:\python_folder\Lib\site-packages\_pytest\main.py", line 243, in _main
17:59:37 INTERNALERROR>     config.hook.pytest_runtestloop(session=session)
17:59:37 INTERNALERROR>   File "D:\python_folder\Lib\site-packages\pluggy\hooks.py", line 284, in __call__
17:59:37 INTERNALERROR>     return self._hookexec(self, self.get_hookimpls(), kwargs)
17:59:37 INTERNALERROR>   File "D:\python_folder\Lib\site-packages\pluggy\manager.py", line 68, in _hookexec
17:59:37 INTERNALERROR>     return self._inner_hookexec(hook, methods, kwargs)
17:59:37 INTERNALERROR>   File "D:\python_folder\Lib\site-packages\pluggy\manager.py", line 62, in <lambda>
17:59:37 INTERNALERROR>     firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
17:59:37 INTERNALERROR>   File "D:\python_folder\Lib\site-packages\pluggy\callers.py", line 203, in _multicall
17:59:37 INTERNALERROR>     gen.send(outcome)
17:59:37 INTERNALERROR>   File "D:\python_folder\Lib\site-packages\pluggy\callers.py", line 81, in get_result
17:59:37 INTERNALERROR>     _reraise(*ex)  # noqa
17:59:37 INTERNALERROR>   File "D:\python_folder\Lib\site-packages\pluggy\callers.py", line 187, in _multicall
17:59:37 INTERNALERROR>     res = hook_impl.function(*args)
17:59:37 INTERNALERROR>   File "D:\python_folder\Lib\site-packages\xdist\dsession.py", line 115, in pytest_runtestloop
17:59:37 INTERNALERROR>     self.loop_once()
17:59:37 INTERNALERROR>   File "D:\python_folder\Lib\site-packages\xdist\dsession.py", line 138, in loop_once
17:59:37 INTERNALERROR>     call(**kwargs)
17:59:37 INTERNALERROR>   File "D:\python_folder\Lib\site-packages\xdist\dsession.py", line 258, in worker_runtest_protocol_complete
17:59:37 INTERNALERROR>     self.sched.mark_test_complete(node, item_index, duration)
17:59:37 INTERNALERROR>   File "D:\python_folder\Lib\site-packages\xdist\scheduler\load.py", line 151, in mark_test_complete
17:59:37 INTERNALERROR>     self.node2pending[node].remove(item_index)
17:59:37 INTERNALERROR> KeyError: <WorkerController gw0>
17:59:47 
17:59:47 ============= 18 failed, 1 passed, 312 warnings in 298.89 seconds =============
```
Above code was executed on jenkins and the one bellow in the console on the same computer.

```
D:\>my_python -m pytest --disable-pytest-warnings -v -l -n1 C:\Users\user\.jenkins\workspace\master_GUI\test\client_test\integration\gui
platform win32 -- Python 2.7.10, pytest-4.1.1, py-1.5.4, pluggy-0.8.1 -- D:\python_folder\python.exe
cachedir: .pytest_cache
rootdir: D:\, inifile:
plugins: xdist-1.26.0, forked-1.0.1, cov-2.5.1
[gw0] win32 Python 2.7.10 cwd: D:\
[gw0] Python 2.7.10 (default, Nov 29 2018, 13:23:51) [MSC v.1900 64 bit (AMD64)]
gw0 [27]

test_component.py::TestComponentEditor::test_component
[gw0] [ 37%] PASSED test_some_other_component.py::test_some_other_component !!!!!!!!!!!!!!!!!!!! D:\python_folder\Lib\site-packages\xdist\workermanage.py:436: Attr
ibuteError: 'module' object has no attribute 'GtkWarning'

INTERNALERROR>   File "D:\python_folder\Lib\site-packages\xdist\workermanage.py", line 336, in process_from_remote
INTERNALERROR>     kwargs["warning_message_data"]
INTERNALERROR>   File "D:\python_folder\Lib\site-packages\xdist\workermanage.py", line 436, in unserialize_warning_message
INTERNALERROR>     cls = getattr(mod, data["message_class_name"])
INTERNALERROR> AttributeError: 'module' object has no attribute 'GtkWarning'
[gw0] node down: D:\python_folder\Lib\site-packages\xdist\workermanage.py:436: AttributeError: 'module' object has no attribute 'GtkWarning'
[gw0] [ 37%] FAILED test_some_other_component.py::test_some_other_component
Replacing crashed worker gw0
!!!!!!!!!!!!!!!!!!!! D:\python_folder\Lib\site-packages\xdist\workermanage.py:436: AttributeError: 'module' object has no attribute 'GtkWarning'
INTERNALERROR>   File "D:\python_folder\Lib\site-packages\xdist\workermanage.py", line 336, in process_from_remote
INTERNALERROR>     kwargs["warning_message_data"]
INTERNALERROR>   File "D:\python_folder\Lib\site-packages\xdist\workermanage.py", line 436, in unserialize_warning_message
INTERNALERROR>     cls = getattr(mod, data["message_class_name"])
INTERNALERROR> AttributeError: 'module' object has no attribute 'GtkWarning'
[gw1] win32 Python 2.7.10 cwd: D:\
gw1 C[gw0] node down: D:\python_folder\Lib\site-packages\xdist\workermanage.py:436: AttributeError: 'module' object has no attribute 'GtkWarning'
Replacing crashed worker gw0
Exception KeyError: (1,) in <function remove at 0x0000000004162208> ignored
[gw2] win32 Python 2.7.10 cwd: D:\
gw2 CINTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "D:\python_folder\Lib\site-packages\_pytest\main.py", line 203, in wrap_session
INTERNALERROR>     session.exitstatus = doit(config, session) or 0
INTERNALERROR>   File "D:\python_folder\Lib\site-packages\_pytest\main.py", line 243, in _main
INTERNALERROR>     config.hook.pytest_runtestloop(session=session)
INTERNALERROR>   File "D:\python_folder\Lib\site-packages\pluggy\hooks.py", line 284, in __call__
INTERNALERROR>     return self._hookexec(self, self.get_hookimpls(), kwargs)
INTERNALERROR>   File "D:\python_folder\Lib\site-packages\pluggy\manager.py", line 68, in _hookexec
INTERNALERROR>   File "D:\python_folder\Lib\site-packages\pluggy\manager.py", line 62, in <lambda>
INTERNALERROR>     firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
INTERNALERROR>   File "D:\python_folder\Lib\site-packages\pluggy\callers.py", line 208, in _multicall
INTERNALERROR>     return outcome.get_result()
INTERNALERROR>   File "D:\python_folder\Lib\site-packages\pluggy\callers.py", line 81, in get_result
INTERNALERROR>     _reraise(*ex)  # noqa
INTERNALERROR>   File "D:\python_folder\Lib\site-packages\pluggy\callers.py", line 187, in _multicall
INTERNALERROR>   File "D:\python_folder\Lib\site-packages\xdist\dsession.py", line 115, in pytest_runtestloop
INTERNALERROR>     self.loop_once()
INTERNALERROR>   File "D:\python_folder\Lib\site-packages\xdist\dsession.py", line 138, in loop_once
INTERNALERROR>     call(**kwargs)
INTERNALERROR>   File "D:\python_folder\Lib\site-packages\xdist\dsession.py", line 208, in worker_errordown
INTERNALERROR>     self._active_nodes.remove(node)
INTERNALERROR> KeyError: <WorkerController gw0>

============ 4 failed, 7 passed, 68736 warnings in 784.27 seconds =============
```

Traceback in jenkins and cmd are equal up to the `line 138, in loop_once` where after they start to diverge but end up with the same `KeyError`.

I'm also using pytest and xdist for running non gui tests and they work perfectly ok.

dprada:
On the controller, `unserialize_warning_message` rebuilt a warning as `cls(*message_args)`, and `serialize_warning_message` sends only `args`. But `args` alone does not describe a warning: the instance keeps state beside them — in its `__dict__`, or in `__slots__` — and we transferred none of it.

So a warning keeping anything beside its `args` arrived without it, and one deriving its message from that state arrived silently wrong. `-n0` serializes nothing and is the reference for what the warning actually says:

```python
class CodeWarning(UserWarning):
    def __init__(self, code=None):
        self.code = code
        super().__init__()
    def __str__(self):
        return "code {} tripped".format(self.code or "unknown")
```

```
-n0:  code 42 tripped
-n1:  code unknown tripped
```

### What this does

**Transfer the state, and rebuild without re-running `__init__`.** `_serializable_warning_state` collects what the instance keeps — its `__dict__` plus the slots along its MRO — and the controller applies it with `cls.__new__(cls)` + `BaseException.__init__` + `__setstate__`, whose `setattr` per key reaches slots as well as the dictionary. `args` and state are each gated on what `execnet.dumps` can carry.

**Read the instance, not `__reduce__`.** The one place where I did not follow the review literally, because `__reduce__` cannot answer the question. It reports only `self.__dict__`, so a `__slots__` class reduces to a two-tuple and its slots read as "keeps nothing": the controller rebuilds an instance whose own `__str__` then raises `AttributeError` where it renders it — an `INTERNALERROR`, not merely a wrong message. In the other direction, a class that replaces `__reduce__` reads as "state we cannot send" even when it keeps nothing, so its class would be dropped and its text prefixed where `master` rebuilds it exactly. Reading the instance makes both exact, and settles the `__reduce__` item without running a callable named by the payload: nothing consults `__reduce__` any more.

**Drop the class only when its state cannot cross.** A warning that keeps nothing has nothing to lose and is rebuilt exactly; one whose state will not serialize takes the existing generic-`Warning` path, for the reason above. That second case carries a deliberate cost: a warning keeping something unserializable but rendering from its `args` alone loses its class, where `master` kept it. Telling those apart means comparing the rendered text across the boundary, which your review rejected — say the word and I will add it as a tie-breaker for exactly that case.

### Tests

Assert the resulting type and the exact text, per the review. Round-trip fidelity for four shapes — field-first, state-only, `__slots__`, and a class that provides its own `__reduce__`; the generic-`Warning` fallback for state that will not serialize; exact rebuild for a warning with no state, so the fallback does not widen to warnings with nothing to lose; and `-n0`/`-n1` parity for the two shapes that used to disagree.

Mutation-verified: reading the state from `__reduce__` instead of from the instance fails 3 tests, dropping the lost-state check 1.

Full suite green. `ruff check`, `ruff format` and `mypy` clean on the changed files.

---

Rebased onto `master` and narrowed to this one question. The guard against a warning's class failing to resolve, which used to be the second half of this PR, is now #1387: it closes #404, and neither half needs the other.

The two red `py311-pytestmain` checks are #1386 (`--dist=loadgroup` silently losing its grouping on pytest main): unrelated to this change, and red on every open PR in the repo since pytest-dev/pytest#14758 merged on 2026-09-06.


RonnyPfannschmidt:
Ill give a more detailed reply when I get back to the computer 

But the premise here is wrong as is

RonnyPfannschmidt (changes requested):
up front: the analysis below and the cell were put together by claude at my direction, and the output block is what it printed running that cell in its own sandbox on cpython 3.12.3. i have not rerun it locally, so treat the numbers as something to check rather than something established. the reasoning and the conclusions are mine. whats put here is iteration 7

---

this misunderstands how warnings and exceptions get reconstructed.

`cls(*args)` is not something xdist made up. [`BaseException.__reduce__`](https://docs.python.org/3/library/exceptions.html#BaseException.args) returns `(cls, self.args)`, plus the instance dict as a third element when non-empty, and pickle calls the class with those args. your first case therefore doubles under plain pickle and copy too, no xdist involved. and [`warnings.warn(msg, category)`](https://docs.python.org/3/library/warnings.html#warnings.warn) builds the instance as `category(msg)`, so neither of your classes can be used as a category at all - one re-renders the message, the other swallows it.

the check in this pr is also at the wrong end. what we get wrong is the state: `__reduce__` carries a third element, `BaseException` implements `__setstate__` (setattr per key, the branch pickle takes for exceptions), and we transfer neither. rebuilding without re-running `__init__` gets both of your cases back exact, class included.

paste this in a cell:

```python
import copy
import pickle
import warnings


class BrokenResourceWarning(UserWarning):
    """field first, message rendered in __init__"""
    def __init__(self, resource):
        self.resource = resource
        super().__init__(f"{resource!r} is not available")


class BrokenCodeWarning(UserWarning):
    """no args at all, text derived from state"""
    def __init__(self, code=None):
        self.code = code
        super().__init__()

    def __str__(self):
        return f"code {self.code or 'unknown'} tripped"


class FixedResourceWarning(UserWarning):
    """message first, structured data kept alongside"""
    def __init__(self, message, resource=None):
        super().__init__(message)
        self.resource = resource

    @classmethod
    def for_resource(cls, resource):
        return cls(f"{resource!r} is not available", resource=resource)


def xdist_today(w):
    """class + args, what unserialize_warning_message does"""
    return type(w)(*w.args)


def via_reduce(w):
    """args plus reduce state, rebuilt without re-running __init__"""
    red = w.__reduce__()
    cls, args = red[0], red[1]
    state = red[2] if len(red) > 2 else None
    new = cls.__new__(cls)
    BaseException.__init__(new, *args)
    if state is not None:
        new.__setstate__(state)
    return new


def report(w):
    print(f"{type(w).__name__}")
    print(f"    {'original':12} {str(w)!r}")
    for name, fn in [
        ("pickle", lambda x: pickle.loads(pickle.dumps(x))),
        ("copy", copy.copy),
        ("xdist today", xdist_today),
        ("via reduce", via_reduce),
    ]:
        try:
            out = repr(str(fn(w)))
        except Exception as exc:
            out = f"{type(exc).__name__}: {exc}"
        print(f"    {name:12} {out}")
    print()


report(BrokenResourceWarning("gpu"))
report(BrokenCodeWarning(42))
report(FixedResourceWarning.for_resource("gpu"))

for cls in (BrokenResourceWarning, BrokenCodeWarning, FixedResourceWarning):
    with warnings.catch_warnings(record=True) as rec:
        warnings.simplefilter("always")
        try:
            warnings.warn("boom", cls)
            got = repr(str(rec[0].message))
        except Exception as exc:
            got = f"{type(exc).__name__}: {exc}"
    print(f"warn('boom', {cls.__name__}) -> {got}")
```

reported output, cpython 3.12.3, sandbox run as noted above:

```
BrokenResourceWarning
    original     "'gpu' is not available"
    pickle       '"\'gpu\' is not available" is not available'
    copy         '"\'gpu\' is not available" is not available'
    xdist today  '"\'gpu\' is not available" is not available'
    via reduce   "'gpu' is not available"

BrokenCodeWarning
    original     'code 42 tripped'
    pickle       'code 42 tripped'
    copy         'code 42 tripped'
    xdist today  'code unknown tripped'
    via reduce   'code 42 tripped'

FixedResourceWarning
    original     "'gpu' is not available"
    pickle       "'gpu' is not available"
    copy         "'gpu' is not available"
    xdist today  "'gpu' is not available"
    via reduce   "'gpu' is not available"

warn('boom', BrokenResourceWarning) -> "'boom' is not available"
warn('boom', BrokenCodeWarning) -> 'code boom tripped'
warn('boom', FixedResourceWarning) -> 'boom'
```

`FixedResourceWarning` is the shape that holds up everywhere: message first, structured datum as a keyword, a classmethod for the convenient call. it survives pickle, copy, the category api and our current serializer unchanged, and needs no xdist patch at all.

what i'd review on our side:

- serialize args and the reduce state, each gated on what `execnet.dumps` can carry
- rebuild via `__new__` + `__setstate__` instead of `cls(*args)`, so a re-rendering `__init__` never runs
- honour a class-provided `__reduce__` when its callable isn't the class itself
- slots state arrives as a `(dict, slots)` 2-tuple - handle it or fall back
- fall back to a plain `Warning` only when the state won't transfer

comparing `str()` and discarding the result detects that our transfer is lossy without making it less lossy, and pays for it by dropping the message class and prefixing the text with `module.Class: `.

tests should assert the resulting type and the exact text, not a substring and an occurrence count.

dprada:
Still reproducible, now differently explained: on `master`, `-n0` and `-n1` disagree about what the same warning says. The first version of this PR misread why, and its check detected the loss instead of repairing it — @RonnyPfannschmidt's review is correct on every point.

Thank you for it. The write-up did more than reject an approach: it explained where the mechanism actually lives, which shape of warning class holds up everywhere, and what you would want to see on your side. I ran your cell on CPython 3.13.14 and it reproduces exactly, `BrokenCodeWarning` included — the case that shows this is not only about badly shaped classes.

I have kept this open rather than closing it because the behaviour it was opened for is still observable, and I believe the five items you listed close it properly. If you conclude otherwise, closing this is entirely reasonable and the analysis you wrote was worth more than the patch either way.

### The defect

`serialize_warning_message` sends only `args`, and `unserialize_warning_message` rebuilds as `cls(*message_args)`. `BaseException.__reduce__` carries the instance dictionary as a third element and `BaseException.__setstate__` applies it; we transfer neither. A warning keeping anything beside its `args` therefore arrives without it, and one deriving its message from that state arrives silently wrong.

`-n0` serializes nothing, so it is the reference for what the warning actually says:

```python
class CodeWarning(UserWarning):
    def __init__(self, code=None):
        self.code = code
        super().__init__()
    def __str__(self):
        return "code {} tripped".format(self.code or "unknown")

def test_func():
    warnings.warn(CodeWarning(42))
```

```
-n0   code 42 tripped
-n1   code unknown tripped
```

That class round-trips correctly under `pickle` and under `copy`. It only breaks crossing a worker boundary, and the result reads like a legitimate message rather than an error.

### The change

Following your five items:

- Serialize the reduce state alongside `args`, each gated on what `execnet.dumps` can carry, so a payload that cannot cross is reported absent rather than half-sent.
- Rebuild with `cls.__new__` + `BaseException.__init__` + `__setstate__`, so a re-rendering `__init__` never runs.
- Handle the `(dict, slots)` two-tuple that `__slots__` classes reduce to.
- Leave a class-provided `__reduce__` alone when its callable is not the class itself, falling back to the generic `Warning`. **This is my reading of your fourth point and may not be what you meant**: `execnet` cannot carry that callable, so I could not find a way to honour such a protocol rather than override it. If you had something else in mind I would rather follow it than guess.
- Fall back to the generic `Warning` only when the state will not transfer.

### Tests

`test_state_beyond_args_survives` is the acceptance test, parametrized over `-n0`/`-n1`; it fails on `master` at `-n1` while passing at `-n0`. Two unit tests cover the shapes from your review — field-first and state-only — asserting the resulting type, the exact text and the restored attribute, rather than a substring and an occurrence count. All three fail without the change.

dprada:
@RonnyPfannschmidt — ready for another look, and I owe you an apology for part of the delay.

The code has implemented your direction since 2026-08-17, but I never rewrote the PR description, so it still argued the first approach — the one you said was at the wrong end. Anyone returning to this PR was reading a design that is no longer in it. That is fixed now; the body above describes what is actually here.

## Since your review

Your five items, and where each landed:

| item | state |
| --- | --- |
| serialize args and the reduce state, each gated on `execnet.dumps` | done |
| rebuild via `__new__` + `__setstate__` | done |
| slots state as a `(dict, slots)` two-tuple | done |
| fall back to a plain `Warning` only when the state won't transfer | **was incomplete — now done** |
| honour a class-provided `__reduce__` when its callable isn't the class | **was missed — see below** |

Two of them were not actually finished when I said the rework was complete, and together they left something worse than the original defect:

```
CustomReduce      worker: "code 7 tripped"  ->  controller: AttributeError: no attribute "code"
UnDumpableState   worker: "code 9 tripped"  ->  controller: AttributeError: no attribute "code"
```

`_serializable_reduce_state` returned `None` both for a warning that carries no state and for one whose state cannot cross, and the controller rebuilt the class either way. Rebuilding without the attributes a class's own `__str__` reads means the controller raises when it renders the warning — a crash introduced while fixing a rendering defect. It returns `(state, lost)` now, and a lost state takes the fallback.

The `__reduce__` item was missed for a reason worth naming: the tuple-length check ran before the callable check, and such a class commonly reduces to a **two-tuple carrying its state inside the args**. That read as "no state, nothing lost". The callable is checked first now.

## One question, on that item

I am not sure I read "honour" correctly. It could mean *rebuild through it*, which is what pickle does. I did not, because it calls an arbitrary callable named by the payload, in the controller's receiver thread — the same place the second half of this PR is about protecting.

What it does instead is treat it as a **loss** rather than an absence, so the class is dropped and the text reported once, instead of an instance arriving broken. If you meant the stronger reading, say so and I will implement it.

## The second half, and #404

Separable — happy to split it into its own PR if you would rather review them apart.

Resolving the warning's class is `importlib.import_module` plus a `getattr`, both outside every guard, both in the receiver thread. This is #404: open since 2019, with your own *"I believe we need some kind of fix"* on it, and its 2022 report names the `getattr` line and shows it arriving as an `INTERNALERROR`. `test_warning_serialization_tweaked_module` was added alongside that report and asserted `pytest.raises(ModuleNotFoundError)` — it characterised the defect rather than fixing it. It now asserts the degradation, and a new test covers the `AttributeError` shape, which is the one that report actually hit.

The category resolution below it had no guard at all and no test, and `category` is what pytest filters and reports on.

## Verification

Rebased onto the current branch head after the master merge and re-run there: **235 passed, 2 skipped, 10 xfailed**. `ruff check`, `ruff format` and `mypy` clean on the changed files.

Each guard is mutation-verified — removing the message-class guard fails 3 tests, the category guard 2, the lost-state check 2 — because a guard nobody has watched fail is not a guard.

dprada:
@RonnyPfannschmidt — rebased, narrowed to one question, and two defects of my own fixed.

**The guard has moved out.** What used to be the second half of this PR is #1387, which closes #404 and needs nothing from here. This PR is now the state question alone.

**Two shapes came out of my rework worse than they went in.** I found them writing tests for shapes I had not covered:

| shape | master | as I had it | now |
| --- | --- | --- | --- |
| `__slots__`, text from a slot | wrong text | **INTERNALERROR** | exact |
| own `__reduce__`, keeps nothing | exact | class dropped, text prefixed | exact |

The cause was reading the state from `__reduce__`. `BaseException.__reduce__` reports only `self.__dict__`, so slots read as "keeps nothing", and the controller rebuilt an instance whose own `__str__` then raised `AttributeError` where it renders it — the run-ending failure the other half exists to prevent, introduced by this one. The `(dict, slots)` two-tuple I had handled turns out to be unreachable: only a class-provided `__reduce__` produces it, and those were rejected one check earlier.

Reading the instance instead — `__dict__` plus the slots along the MRO — makes both exact, and **moots the question I asked you about "honour"**: nothing consults `__reduce__` any more, so there is no callable to decide about.

**One residual cost, stated in the description.** A warning that keeps unserializable state but renders from its `args` alone now loses its class, where `master` kept it. That follows your fifth item literally. Telling that case apart means comparing the rendered text across the boundary — the mechanism you rejected — so I did not add it on my own; say the word and I will.

PR #1382, opened 2026-09-10
Fix loadscope handling of colons in parameters

### LINKED ISSUE #1335: loadscope bug: `::` in test parameters breaks scope detection
## Summary

`LoadScopeScheduling._split_scope()` uses `nodeid.rsplit("::", 1)` to
determine the test scope (file).  When a parametrized test value contains
`::` (e.g. IPv6 addresses), the split lands inside the parameter instead
of at the `.py::` boundary.  Tests from the same file get assigned to
different workers, each paying the full fixture setup/teardown cost.

## Affected code

`xdist/scheduler/loadscope.py`, `_split_scope()`:

```python
def _split_scope(self, nodeid: str) -> str:
    return nodeid.rsplit("::", 1)[0]
```

## Example

```
nodeid: synthrecord/tests_synthrecord.py::test_forward[cafe:cafe::cafe-AAAA]
                                                             ^^
                                                      rsplit lands here

expected scope: synthrecord/tests_synthrecord.py
actual scope:   synthrecord/tests_synthrecord.py::test_forward[cafe:cafe
```

## Impact observed in BIND 9 CI

Six `synthrecord` tests have IPv6 addresses (`::1`, `cafe:cafe::cafe`,
`cafe::`) in their parameters.  These get misclassified into separate
scopes and routed to different workers.  Each triggers a full `named`
startup/shutdown cycle (~30 s).  In a real CI run ([job 7468217](https://gitlab.isc.org/isc-projects/bind9/-/jobs/7468217/raw)), the
synthrecord scope spanned 1001 s instead of ~8 s because the orphaned
tests were interleaved with unrelated scopes on other workers.

## Suggested fix

Split on `.py::` instead of the last `::`:

```python
def _split_scope(self, nodeid: str) -> str:
    if ".py::" in nodeid:
        return nodeid.split(".py::")[0] + ".py"
    return nodeid.rsplit("::", 1)[0]
```

This is unambiguous: `.py::` only appears at the module boundary.

## Reproduction

```python
from xdist.scheduler.loadscope import LoadScopeScheduling

s = LoadScopeScheduling.__new__(LoadScopeScheduling)

# Normal test — correct scope
print(s._split_scope("foo/bar.py::test_one"))
# -> foo/bar.py

# IPv6 in parameter — wrong scope
print(s._split_scope("foo/bar.py::test_ip[::1-expected]"))
# -> foo/bar.py::test_ip[
```

LouisDeconinck:
Fixes #1335.

`LoadScopeScheduling._split_scope` previously split on the last `::`, so a parametrized value containing `::` could be mistaken for the test scope. Parse the module boundary first and only consider `::` separators before parametrization, preserving class-level grouping.

Added regression coverage for colon-containing parameters and parameterized class methods.

Checks:
- [x] Reasonable regression tests added
- [x] Towncrier changelog fragment added
- [x] `pytest -q testing` (224 passed, 6 skipped, 10 xfailed)
- [x] `ruff check` and `ruff format --check`
- [x] `git diff --check`


LouisDeconinck:
Pushed a fix for the `py311-pytestmain` CI failures.

The failing `TestGroupScope` tests were unrelated to the loadscope change: pytest main (pytest-dev/pytest#14758, upcoming 9.2) replaced the `Node._nodeid` string with a structured `NodeId` object in `Node._id`, so `item._nodeid = ...` in `remote.py` silently wrote a dead attribute and the `@group` suffix never reached the scheduler — every test became its own scope.

The commit detects `Node._id` and overrides the `NodeId`'s cached string form instead, so `item.nodeid` (and reports, which share the object) still carry the `@group` suffix; released pytest keeps the existing `_nodeid` path. Full `testing/` suite passes locally against both pytest main and pytest 9.1.1.

Note this affects `master` too — happy to split it into a separate PR if preferred.

PR #1388, opened 2026-09-21
Fix loadgroup with pytest main node IDs

### LINKED ISSUE #1386: `--dist=loadgroup` silently stops grouping on pytest main: the nodeid mutation no longer takes effect
On pytest `main`, `--dist=loadgroup` no longer groups tests by `xdist_group`. Nothing fails: the tests are simply distributed as if the mark were not there.

## Reproducer

`test_a.py`, and an identical `test_b.py`:

```python
import pytest

class TestA:
    @pytest.mark.xdist_group(name="xdist_group")
    @pytest.mark.parametrize("i", range(5))
    def test(self, i):
        pass
```

```
pytest -n2 --dist=loadgroup -v
```

xdist master (`eba6a44`) with pytest 9.1.1 — one group, one worker:

```
[gw0] [ 10%] PASSED test_a.py::TestA::test[0]@xdist_group
[gw0] [ 20%] PASSED test_a.py::TestA::test[1]@xdist_group
[gw0] [ 30%] PASSED test_a.py::TestA::test[2]@xdist_group
[gw0] [ 40%] PASSED test_a.py::TestA::test[3]@xdist_group
[gw0] [ 50%] PASSED test_a.py::TestA::test[4]@xdist_group
[gw0] [ 60%] PASSED test_b.py::TestA::test[0]@xdist_group
[gw0] [ 70%] PASSED test_b.py::TestA::test[1]@xdist_group
[gw0] [ 80%] PASSED test_b.py::TestA::test[2]@xdist_group
[gw0] [ 90%] PASSED test_b.py::TestA::test[3]@xdist_group
[gw0] [100%] PASSED test_b.py::TestA::test[4]@xdist_group
```

The same xdist with pytest 9.2.0.dev340+g6a0de9be5 — no suffix, round-robin:

```
[gw1] [ 10%] PASSED test_a.py::TestA::test[1]
[gw0] [ 20%] PASSED test_a.py::TestA::test[0]
[gw1] [ 30%] PASSED test_a.py::TestA::test[3]
[gw0] [ 40%] PASSED test_a.py::TestA::test[2]
[gw0] [ 50%] PASSED test_b.py::TestA::test[0]
[gw1] [ 60%] PASSED test_a.py::TestA::test[4]
[gw0] [ 70%] PASSED test_b.py::TestA::test[1]
[gw1] [ 80%] PASSED test_b.py::TestA::test[2]
[gw1] [ 90%] PASSED test_b.py::TestA::test[4]
[gw0] [100%] PASSED test_b.py::TestA::test[3]
```

## Cause

The group name reaches the controller as a suffix on the nodeid, written in `src/xdist/remote.py:264`:

```python
item._nodeid = f"{item.nodeid}@{'_'.join(sorted(gnames))}"
```

pytest-dev/pytest#14758 (`431f3e1f5`, merged 2026-09-06) replaced string nodeids with a structured `NodeId`. `Node.nodeid` is now a read-only property over `self._id`, so that line writes an attribute nobody reads — silently, without raising.

With no `@` in the nodeid, `LoadGroupScheduling._split_scope` returns the nodeid itself:

```python
if nodeid.rfind("@") > nodeid.rfind("]"):
    return nodeid.split("@")[-1]
else:
    return nodeid
```

so every test becomes its own scope. That is why the tests scatter rather than falling back to module or class grouping.

For what it is worth, that pytest PR kept the public `nodeid` string as a backward-compatible read surface; what breaks here is our *write* to the private `_nodeid`.

## Where this shows up

`py311-pytestmain` is red on every open PR — #1383, #1384 and #1385 (run from today) — always the same five tests: `TestGroupScope::test_by_module`, `test_by_class`, `test_module_single_start`, `test_multiple_group_marks` and `test_multiple_group_order`.

master still looks unaffected only because its test workflow last ran `pytestmain` on 2026-09-04, two days before that pytest merge.

No released pytest is affected, so nobody is broken today. When 9.2 ships, tests that need to share a worker will quietly stop doing so, which is the kind of failure that is hard to attribute from the outside.

## Related

#1119 ("Loadgroup: Don't modify test nodeids any longer", Aug 2024) proposed replacing this side-channel with an explicit parameter. I have no view on whether that is the right fix now versus adapting the write — just noting that it exists and is still open.

## Environment

- pytest-xdist master `eba6a44`
- pytest `9.2.0.dev340+g6a0de9be5` (and 9.1.1 for the comparison)
- execnet 2.1.2, CPython 3.13.14, Linux

itscloud0:
## Problem

pytest main uses structured node IDs. `--dist=loadgroup` still writes the legacy `_nodeid` field, so grouped tests silently lose their group suffix and are distributed independently.

## Solution

Update the structured ID when it is available, while retaining the legacy path for supported older pytest releases.

## Verification

- `pytest testing/acceptance_test.py::TestGroupScope -q` (6 passed)
- `ruff check --select B010 src/xdist/remote.py`
- `ruff format --check src/xdist/remote.py`
- `git diff --check`

Closes #1386.


RonnyPfannschmidt:
@nicoddemus i tihnk we need to transfer the groups separately -  id was never supposed to be mutated to begin wirth

